PR #18, opened 2020-08-25
test(lib) add unit tests

lucasfernog:


src/lib.rs
```diff
@@ -145,15 +190,128 @@ mod test_utils {
...
+
+        if test_type == TestType::Storage || test_type == TestType::SigningAndStorage {
+            // ---- Stronghold storage ----
+            #[cfg(feature = "stronghold-storage")]
+            test_cases.push(ManagerTestCase::Storage(StorageTestCase {
+                storage_password: None,
+                storage: ManagerStorage::Stronghold,
+            }));
+
+            test_cases.push(ManagerTestCase::Storage(StorageTestCase {
```
rajivshah3: Is this specific to `stronghold-storage` as well?

lucasfernog: yeah you're right, thanks for finding that :P

src/account/mod.rs
```diff
@@ -640,35 +640,364 @@ impl Account {
...
+                assert_eq!(
+                    account_handle.read().await.addresses(),
+                    &vec![spent_address, unspent_address1, unspent_address2]
+                );
+            },
+        )
+        .await;
+    }
+
+    // TODO list_addresses tests
```
Thoralf-M: Remove this comment?

src/client.rs
```diff
@@ -254,3 +254,158 @@ pub struct ClientOptions {
...
+    #[test]
+    fn multi_node_constructor() {
+        let nodes = ["https://tangle.iotaqubic.us:14267"];
+        let quorum_size = 5;
+        let quorum_threshold = 0.5;
+        let client = ClientOptionsBuilder::nodes(&nodes)
+            .unwrap()
+            .quorum_size(quorum_size)
+            .quorum_threshold(quorum_threshold)
+            .build()
```
Thoralf-M, line 320: Should the builder return an error if there is only one node, but quorum_size = 5?

Thoralf-M:
Maybe you should use `https://api.lb-0.testnet.chrysalis2.com` as URL instead of the one with iota1.0 nodes

src/account/mod.rs
bingyanglin: Yes please.

src/client.rs
bingyanglin: I think it's good to have error returning for each function.

src/account/mod.rs
lucasfernog: thanks :)

src/client.rs
lucasfernog: Since iota.rs removed quorum, I'm removing this on another PR too.

Thoralf-M (approved):

Thoralf-M (approved):

PR #33, opened 2020-08-27
fix(storage) wrap adapter in a Mutex to prevent threading errors

lucasfernog:
# Description of change

The stronghold adapter wasn't passing tests because of the usage across threads and a file lock issue. This PR resolves that by wrapping the adapter in a [Mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html). 

## Type of change

By running the unit tests.

## Change checklist


Cargo.toml
```diff
@@ -33,4 +30,4 @@ tokio = { version = "0.2", features = ["full"] }
 
 [features]
 default = ["vault", "crypto", "random", "snapshot", "primitives", "bincode"]
-sqlite = ["rusqlite"]
+sqlite = []
```
wusyong, line 33: I think it's still better to add rusqlite here and add attributes to sqlite module.

Cargo.toml
```diff
@@ -11,14 +11,11 @@ getset = "0.1"
...
 once_cell = "1.4"
-kv = "0.20"
 iota-core = { git = "https://github.com/iotaledger/iota.rs", branch = "dev" }
 url = { version = "2.1", features = [ "serde" ] }
 tokio = "0.2"
 rand = "0.3"
-
-# sql
-rusqlite = { version = "0.23", optional = true }
+rusqlite = "0.23"
```
wusyong, line 18: What do you think about some other choice like `sqlx`? It can also support different databases.

lucasfernog: to use `sqlx` here I'd need to change the StorageAdapter to make every method `async`, but I like the idea and I'll do it on a separate PR.

wusyong (approved):

PR #48, opened 2020-09-22
refactor(lib) migration for chrysalis part 2 changes

lucasfernog:
# Description of change

This PR updates the library code to use the `chrysalis` branch of the client library dependency, applying the chrysalis part 2 changes. 

## Type of change

- Refactor

Running unit tests.

## Change checklist

### DIFF
--- a/src/account/sync/input_selection.rs
+++ b/src/account/sync/input_selection.rs
@@ -107,7 +107,7 @@ mod tests {
         for i in 0..utxos_number {
             available_utxos.push(
                 AddressBuilder::new()
-                    .address(IotaAddress::from_ed25519_bytes([0; 32]))
+                    .address(IotaAddress::from_ed25519_bytes(&[0; 32]))
                     .balance(rng.gen_range(0, 2000))
                     .key_index(i)
                     .build()
--- a/src/account/sync/input_selection.rs
+++ b/src/account/sync/input_selection.rs
@@ -1,10 +1,7 @@
 use crate::address::Address;
 use rand::{thread_rng, Rng};
 
-pub fn select_input(
-    target: u64,
-    available_utxos: &mut Vec<Address>,
-) -> crate::Result<Vec<Address>> {
+pub fn select_input(target: u64, available_utxos: &mut [Address]) -> crate::Result<Vec<Address>> {
     if target
         > available_utxos
             .iter()
@@ -25,10 +22,7 @@ pub fn select_input(
     }
 }
 
-fn single_random_draw(
-    target: u64,
-    available_utxos: &mut Vec<Address>,
-) -> crate::Result<Vec<Address>> {
+fn single_random_draw(target: u64, available_utxos: &mut [Address]) -> crate::Result<Vec<Address>> {
     thread_rng().shuffle(available_utxos);
     let mut sum = 0;
 
@@ -48,7 +42,7 @@ fn single_random_draw(
 
 fn branch_and_bound(
     target: u64,
-    available_utxos: &mut Vec<Address>,
+    available_utxos: &mut [Address],
     depth: usize,
     current_selection: &mut Vec<Address>,
     effective_value: u64,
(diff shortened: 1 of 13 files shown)


src/message.rs
```diff
@@ -135,69 +128,49 @@ impl Value {
...
-            trunk_transaction: *tx.trunk(),
-            branch_transaction: *tx.branch(),
-            bundle_hash: *tx.bundle(),
-            nonce: "TX NONCE".to_string(), // TODO
+impl Message {
+    pub(crate) fn from_iota_message(message: IotaMessage) -> crate::Result<Self> {
+        let message = Self {
+            version: 1,
+            trunk: message.trunk().clone(),
+            branch: message.branch().clone(),
```
wusyong, line 160: trunk and branch fields are public iirc. So you could just move them instead of cloning.

src/message.rs
```diff
@@ -207,25 +180,48 @@ impl Transaction {
...
+                .iter()
+                .fold(0, |acc, output| acc + output.amount()),
+            _ => 0,
+        };
+        Value::new(amount.try_into().unwrap(), ValueUnit::I)
+    }
 }
 
-impl PartialEq for Transaction {
+impl PartialEq for Message {
```
wusyong, line 208: `bee_transaction::Message` has this trait, so it should be fine to derive it. (and maybe derive other traits too)

src/address.rs
```diff
@@ -129,40 +119,21 @@ pub(crate) async fn get_addresses(account: &Account, count: usize) -> crate::Res
...
-    for _ in 1..10 {
-        if let Some(trit) = hash.pop() {
-            trits.push(trit);
-        } else {
-            return Err(anyhow::anyhow!("Hash error"));
-        }
-    }
-
-    Ok(TritBuf::from_trits(&trits[..]))
+pub(crate) fn generate_checksum(address: &IotaAddress) -> crate::Result<String> {
```
wusyong, line 122: I think there's no need for the checksum. We are using bech32 address string from now on. This can be generated from method of `bee_transaction::Address`.

src/account/sync/input_selection.rs
```diff
@@ -1,11 +1,19 @@
+use crate::address::Address;
 use rand::{thread_rng, Rng};
 
-pub fn select_input(target: u64, available_utxos: &mut Vec<u64>) -> crate::Result<Vec<u64>> {
-    if target > available_utxos.iter().sum::<u64>() {
+pub fn select_input(
+    target: u64,
+    available_utxos: &mut Vec<Address>,
```
wusyong, line 6:
```suggestion
    available_utxos: &mut [Address],
```

I would prefer slice since `Vec` has `Deref` trait and could take advantage of deref coercion. And other type that has slice as their Deref output could also benefit from it.

src/account/sync/input_selection.rs
```diff
@@ -17,29 +25,32 @@ pub fn select_input(target: u64, available_utxos: &mut Vec<u64>) -> crate::Resul
     }
 }
 
-fn single_random_draw(target: u64, available_utxos: &mut Vec<u64>) -> crate::Result<Vec<u64>> {
+fn single_random_draw(
+    target: u64,
+    available_utxos: &mut Vec<Address>,
```
wusyong, line 30:
```suggestion
    available_utxos: &mut [Address],
```

Ditto, and probably all parameters with `&mut Vec<Address>` below.

src/address.rs
lucasfernog: hmm so bech32 doesn't need address checksum because the chance of it being incorrect is minimal? cc @laumair

wusyong (approved):

src/address.rs
laumair: > hmm so bech32 doesn't need address checksum because the chance of it being incorrect is minimal?

@lucasfernog @wusyong  I am not sure I follow. Bech32 addresses do have a checksum for the HRP (Human Readable Part).

PR #70, opened 2020-10-23
refactor(storage) enable different storage instances per AccountManager

lucasfernog:
# Description of change

In the future, the wallet will need to support management of multiple stronghold snapshots. In order to do that, I've removed the current `OnceCell` StorageAdapter global singleton, and made it a HashMap keyed by the path to the storage file. The key is held by an AccountManager instance and used across the library to get access to the storage adapter.

## Type of change

- Refactor

Unit tests.

## Change checklist

### DIFF
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -20,46 +20,73 @@ use stronghold::Stronghold;
 ///
 /// Used to manage multiple accounts.
 #[derive(Default)]
-pub struct AccountManager {}
+pub struct AccountManager {
+    storage_path: PathBuf,
+}
 
 fn mutate_account_transaction<F: FnOnce(&Account, &mut Vec<Message>)>(
+    storage_path: &PathBuf,
     account_id: AccountIdentifier,
     handler: F,
 ) -> crate::Result<()> {
-    let mut account = crate::storage::get_account(account_id)?;
+    let mut account = crate::storage::get_account(&storage_path, account_id)?;
     let mut transactions: Vec<Message> = account.messages().to_vec();
     handler(&account, &mut transactions);
     account.set_messages(transactions);
-    let adapter = crate::storage::get_adapter()?;
-    adapter.set(account_id, serde_json::to_string(&account)?)?;
+    crate::storage::with_adapter(&storage_path, |storage| {
+        storage.set(account_id, serde_json::to_string(&account)?)
+    })?;
     Ok(())
 }
 
 impl AccountManager {
     /// Initialises a new instance of the account manager with the default storage adapter.
-    pub fn new() -> Self {
-        Default::default()
+    pub fn new() -> crate::Result<Self> {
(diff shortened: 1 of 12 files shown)


src/storage/mod.rs
```diff
@@ -6,53 +6,39 @@ mod stronghold;
...
 pub(crate) fn stronghold_snapshot_filename() -> &'static str {
     "snapshot"
 }
 
-pub(crate) fn get_storage_path() -> &'static PathBuf {
-    STORAGE_PATH.get_or_init(|| "./example-database".into())
-}
-
 /// gets the storage adapter
 #[allow(clippy::borrowed_box)]
```
wusyong, line 31: Do you still need this attribute?

src/account_manager.rs
```diff
@@ -20,46 +20,73 @@ use stronghold::Stronghold;
...
+    })?;
     Ok(())
 }
 
 impl AccountManager {
     /// Initialises a new instance of the account manager with the default storage adapter.
-    pub fn new() -> Self {
-        Default::default()
+    pub fn new() -> crate::Result<Self> {
+        Self::with_storage_path("./example-database")
```
wusyong, line 45: If this is temporary, maybe create a constant string would be better.

src/account_manager.rs
```diff
@@ -20,46 +20,73 @@ use stronghold::Stronghold;
...
+            .storage_path
+            .join(crate::storage::stronghold_snapshot_filename());
         let stronghold = Stronghold::new(
             &stronghold_path,
             !stronghold_path.exists(),
             password.as_ref().to_string(),
             None,
         )?;
-        crate::init_stronghold(stronghold_path, stronghold);
+        crate::init_stronghold(self.storage_path.clone(), stronghold);
```
wusyong, line 79: Maybe `init_stronghold` should take the reference instead.

src/account_manager.rs
```diff
@@ -20,46 +20,73 @@ use stronghold::Stronghold;
...
         Ok(())
     }
 
     /// Enables syncing through node events.
-    pub fn sync_through_events(&self) {
+    pub fn sync_through_events(&'static self) {
```
wusyong, line 84: I'm not sure about static lifetime here. Are you going to create an `AccountManager` with OnceCell or something similar?

src/storage/mod.rs
lucasfernog: yeah I don't need it anymore, good catch, forgot about it

src/account_manager.rs
lucasfernog: pushed a change to make it clone the storage_path instead of moving `self` to a closure, what do you think?

src/account_manager.rs
lucasfernog: you're right

wusyong (approved):

PR #119, opened 2020-11-16
fix(lib) compilation issues

lucasfernog:
# Description of change

Some `bee` changes resulted in compilation issues on the lib. 

## Type of change

### DIFF
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -607,9 +607,10 @@ mod tests {
                     .with_parent1(MessageId::new([0; 32]))
                     .with_parent2(MessageId::new([0; 32]))
                     .with_payload(Payload::Indexation(Box::new(Indexation::new(
-                        "".to_string(),
+                        "index".to_string(),
                         &[0; 16],
                     ).unwrap())))
+                    .with_network_id(0)
                     .finish()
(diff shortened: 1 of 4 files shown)


src/account_manager.rs
```diff
@@ -608,8 +608,8 @@ mod tests {
                     .with_parent2(MessageId::new([0; 32]))
                     .with_payload(Payload::Indexation(Box::new(Indexation::new(
                         "".to_string(),
```
rajivshah3, line 610: I think the test is failing because empty indexes are no longer allowed (https://github.com/Alex6323/bee-p/blob/1ad65ed0909bf2cc55c3b61e16d7bf0b1f124f7f/bee-message/src/payload/indexation.rs#L35)

lucasfernog: yeah I realised that on the tests branch but forgot to add it here, thanks!

rajivshah3: I changed `"".to_string()` to `"a".to_string()` and now it's failing in another place: 

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: MissingField("network_id")', src/account_manager.rs:614:22
```

rajivshah3 (approved):

PR #122, opened 2020-11-18
refactor(event): payloads with Message instead of MessageId, remove clone

lucasfernog:
# Description of change

Changes the event payloads to take references to remove `clone()`s and changes the payloads to include `Message` instead of `MessageId` to simplify usage by consumers. 

## Type of change

Manually with the frontend.

## Change checklist

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -1,5 +1,5 @@
 use crate::address::Address;
-use iota::message::prelude::MessageId;
+use crate::message::Message;
 
 use getset::Getters;
 use once_cell::sync::Lazy;
@@ -10,12 +10,12 @@ use std::sync::{Arc, Mutex};
 /// The balance change event data.
 #[derive(Getters, Serialize)]
 #[getset(get = "pub")]
-pub struct BalanceEvent {
+pub struct BalanceEvent<'a> {
     /// The associated account identifier.
     #[serde(rename = "accountId")]
-    account_id: [u8; 32],
+    account_id: &'a [u8; 32],
     /// The associated address.
-    address: Address,
+    address: &'a Address,
     /// The new balance.
     balance: u64,
 }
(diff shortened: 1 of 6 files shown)


src/event.rs
```diff
@@ -105,7 +103,7 @@ fn error_listeners() -> &'static ErrorListeners {
 }
 
 /// Listen to balance changes.
-pub fn on_balance_change<F: Fn(BalanceEvent) + Send + 'static>(cb: F) {
+pub fn on_balance_change<F: Fn(&BalanceEvent<'_>) + Send + 'static>(cb: F) {
     let mut l = balance_listeners()
```
bingyanglin, line 107: Do we really need to use Mutex for the listeners in the non async functions?

lucasfernog: I'm wrapping the listeners in a Mutex because the listeners are stored in a mutable vector (but it could also be a RwLock)

bingyanglin: Yes I mean do we need it? Because on_balance_change here it is not an async function and will be executed sequentially. So would like to know if there is a scenario that there are more than 1 location the listeners are read/written.

bingyanglin: (simultaneously I mean)

lucasfernog: it's always sequentially but it might be accessed by multiple threads on consumers of the library. but we mainly use it to make the Vec<Listeners> mutable.

bingyanglin: Got it. If multiple threads need to access the listeners in the same time then we need `Mutex` (or RwLock) here. LGTM.

bingyanglin (approved):

bingyanglin (approved):

bingyanglin (approved):

bingyanglin (approved):

cvarley100 (approved):

PR #123, opened 2020-11-18
feat(account): implement retry, reattach and promote, closes #115

### LINKED ISSUE #115: Implementation of promotion and reattachment methods
## Description

The library currently misses the implementation of promotion and reattachment processes. They should be implemented. 

## Motivation

Promotion or reattachment will be required (although very rarely) to speed up the confirmation times of a message. The library should support it for the users to manually reattach or promote. It should also have the ability to automatically reattach or promote messages. 

## Requirements

The implementation should probably leverage iota.rs — see https://github.com/iotaledger/iota.rs/issues/151. 

The specification might also require some updates @laumair. 

Yes.

lucasfernog:
# Description of change

`retry`, `reattach` and `promote` APIs. 

#115 

## Type of change

Not tested yet.

## Change checklist

### DIFF
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -279,10 +279,21 @@ impl Account {
         from: usize,
         message_type: Option<MessageType>,
     ) -> Vec<&Message> {
+        let mut found_message_payloads = vec![];
         let messages_iter = self
             .messages
             .iter()
+            // iterate in reverse order to find reattachments first
+            .rev()
             .filter(|message| {
+                // ignore if the message if one with the same payload was already found (i.e. this message was reattached)
+                if found_message_payloads
+                    .iter()
+                    .any(|payload| payload == &message.payload())
+                {
+                    return false;
+                }
+                found_message_payloads.push(message.payload());
                 if let Some(message_type) = message_type.clone() {
                     match message_type {
                         MessageType::Received => *message.incoming(),
@@ -295,6 +306,8 @@ impl Account {
                     true
                 }
             })
+            // reverse the iterator again to list in the ascending order
+            .rev()
             .skip(from);
         if count == 0 {
             messages_iter.collect()
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -279,36 +279,39 @@ impl Account {
         from: usize,
         message_type: Option<MessageType>,
     ) -> Vec<&Message> {
-        let mut found_message_payloads = vec![];
-        let messages_iter = self
-            .messages
-            .iter()
-            // iterate in reverse order to find reattachments first
-            .rev()
-            .filter(|message| {
-                // ignore if the message if one with the same payload was already found (i.e. this message was reattached)
-                if found_message_payloads
-                    .iter()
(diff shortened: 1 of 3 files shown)


src/account/mod.rs
```diff
@@ -279,10 +279,21 @@ impl Account {
...
         let messages_iter = self
             .messages
             .iter()
+            // iterate in reverse order to find reattachments first
+            .rev()
             .filter(|message| {
+                // ignore if the message if one with the same payload was already found (i.e. this message was reattached)
+                if found_message_payloads
+                    .iter()
+                    .any(|payload| payload == &message.payload())
```
laumair, line 292: @lucasfernog It is possible (though less probable) that an original message gets confirmed even after it was reattached. Considering this, we should always return the confirmed one. So when we find a reattachment, we should check if any of the message is confirmed and then give precedence to the confirmed one.

src/account/sync/mod.rs
```diff
@@ -539,14 +539,102 @@ impl SyncedAccount {
...
+}
+
+async fn repost_message(
+    account_id: AccountIdentifier,
+    storage_path: &PathBuf,
+    message_id: &MessageId,
+    action: RepostAction,
+) -> crate::Result<Message> {
+    let mut account: Account = crate::storage::get_account(&storage_path, account_id)?;
+    match account.get_message(message_id) {
```
laumair, line 589: @lucasfernog At this point, we also need to get the reattachments for this message from the database. We need to check metadata for all messages (original + reattachments). It's possible that one of them got already confirmed and we are looking for metadata for the unconfirmed one.

laumair (changes requested):

laumair:
@lucasfernog We need to also include `retry()` to poll cycle. Also, we need to make sure we do not keep retrying such messages that already have one of their reattachments confirmed.

bingyanglin:
Agree with @laumair, we might need to poll and monitor the all reattachments are confirmed or not, if one of them is already confirmed then we should not `retry()` again.

src/account/mod.rs
lucasfernog: fixed on this [commit](https://github.com/iotaledger/wallet.rs/pull/123/commits/63ebcb2d9b68e283ca945cc5210be06088ff6802)

src/account/sync/mod.rs
lucasfernog: I'm using the `list_messages` algorithm [here](https://github.com/iotaledger/wallet.rs/blob/63ebcb2d9b68e283ca945cc5210be06088ff6802/src/account/sync/mod.rs#L593) to get the actual message we care about and if it's confirmed, we error out

lucasfernog:
`retry()` is part of the polling [here](https://github.com/iotaledger/wallet.rs/blob/65a0351bcf0ef8cf9896cfbc108be4c3a24fbf74/src/account_manager.rs#L422). It leverages the `list_accounts` method so only the actual unconfirmed reattachment will be reattached again if necessary.

laumair (approved):

bingyanglin (approved):

PR #130, opened 2020-11-25
feat(lib): add actor interface crate

lucasfernog:
# Description of change

Adds a crate wrapping the wallet library as an actor. 

## Type of change

With Firefly.

## Change checklist


src/actor/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+    rx: UnboundedReceiver<Message>,
+    message_handler: WalletMessageHandler,
+}
+
+impl Wallet {
+    /// Runs the actor.
+    pub async fn run(mut self) {
+        println!("running wallet actor");
+
+        while let Some(message) = self.rx.recv().await {
```
bingyanglin: It might be good to receive a shutdown signal here to shutdown the `Wallet`, or it might also be good to provide an example/test case to drop the `UnboundedSender` channel, so the user can follow it easily and shutdown the `Wallet` correctly.

lucasfernog: moved this to an example/testcase since it's opinionated and users might use their own actor structure to use the WalletMessageHandler (like firefly uses riker's actor struct instead of this Wallet one).

bingyanglin (approved):
LGTM

PR #149, opened 2020-12-08
feat(bindings): add option to change storage path on nodejs

lucasfernog:
# Description of change

Option to use SQLite as storage adapter on nodejs.

## Type of change

With the nodejs unit test.

## Change checklist

### DIFF
--- a/bindings/node/lib/index.d.ts
+++ b/bindings/node/lib/index.d.ts
@@ -109,6 +109,16 @@ export declare interface AccountToCreate {
   createdAt?: string;
 }
 
+export declare enum StorageType {
+  Stronghold = 1,
+  Sqlite = 2
+}
+
+export declare interface ManagerOptions {
+  storagePath?: string
+  storageType?: StorageType
+}
+
 export declare class AccountManager {
   constructor(storagePath?: string)
   setStrongholdPassword(password: string): void
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -10,14 +10,21 @@
 // See the License for the specific language governing permissions and limitations under the License.
 
 use super::JsAccount;
-use std::sync::{Arc, RwLock};
+use std::{
+    path::PathBuf,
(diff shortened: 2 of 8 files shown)


bindings/node/lib/index.d.ts
```diff
@@ -109,6 +109,16 @@ export declare interface AccountToCreate {
   createdAt?: string;
 }
 
+export declare enum StorageType {
+  Stronghold = 1,
```
thibault-martinez, line 113: Any reason not to start at 0 :) ?

bindings/node/native/src/classes/account_manager/mod.rs
```diff
@@ -50,18 +57,44 @@ fn js_value_to_account_id(
...
                 }
-                None => None,
+                None => Default::default(),
             };
-            let manager = match storage_path {
-                Some(p) => AccountManager::with_storage_path(p),
-                None => AccountManager::new(),
+            let manager = match options.storage_type {
+                StorageType::Sqlite => AccountManager::with_storage_adapter(&options.storage_path, SqliteStorageAdapter::new(&options.storage_path, "accounts").unwrap()),
+                StorageType::Stronghold => AccountManager::with_storage_adapter(&options.storage_path, StrongholdStorageAdapter::new(&options.storage_path).unwrap()),
```
thibault-martinez, line 97: Are these two unwraps fine ? Without wallet.rs background I can't tell.
As a general rule, try to always justify unwraps with a comment, it makes it easier to review/understand.

thibault-martinez (commented):
LGTM

bindings/node/lib/index.d.ts
lucasfernog: no reason really, I could change it later (there's another enum starting at 1)

bindings/node/native/src/classes/account_manager/mod.rs
lucasfernog: unwrap on the binding is fine because it ends up being a thrown exception on the JS side

thibault-martinez (approved):

PR #154, opened 2020-12-10
fix(transfer): lock selected addresses so concurrent transfers can work

lucasfernog:
# Description of change

Currently, sending two transfers concurrently makes the second transaction conflict since the same addresses are selected for both. Locking the selected input addresses prevent that, although if there's no alternative address to select, one of the transfers will fail with a "insufficient funds" error. Ideally, the transfer would be queued to wait for the first message's resolution, but we'll implement such mechanism in the future.

## Type of change

Running this example:

```rust
use iota_wallet::{
    account::{Account, SyncedAccount},
    account_manager::AccountManager,
    client::ClientOptionsBuilder,
    message::Transfer,
};

async fn transfer_(sync_accounts: &Vec<SyncedAccount>, account: &Account) {
    let sync_account = sync_accounts.first().unwrap();
    let data = sync_account
        .transfer(Transfer::new(account.latest_address().unwrap().address().clone(), 150))
        .await;
    println!("{:?}", data);
}

#[tokio::main]
async fn main() -> iota_wallet::Result<()> {
    let mut manager = AccountManager::new().unwrap();
    manager.set_stronghold_password("password").unwrap();

    // first we'll create an example account and store it
    let client_options = ClientOptionsBuilder::node("http://localhost:14265")?.build();
    let account = manager.create_account(client_options).alias("alias").mnemonic("<mnemonic seed with funds here>").initialise()?;

    // we need to synchronize with the Tangle first
    let sync_accounts = manager.sync_accounts().await?;
    let account = manager.get_account(account.id().into()).unwrap();

    let mut handles = vec![];
    for _ in 1..3 {
        let sync_accounts = sync_accounts.clone();
        let account = account.clone();
        let handle = std::thread::spawn(move || {
            let mut runtime = tokio::runtime::Runtime::new().unwrap();
            runtime.block_on(transfer_(&sync_accounts, &account));
        });
        handles.push(handle);
    }

    for handle in handles {
        handle.join().unwrap();
    }

    Ok(())
}
```

## Change checklist

### DIFF
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    address::Address,
+    address::{Address, IotaAddress},
     client::ClientOptions,
     message::{Message, MessageType},
     signing::{with_signer, SignerType},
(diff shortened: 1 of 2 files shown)


laumair (approved):

src/account/mod.rs
```diff
@@ -11,14 +11,32 @@ use crate::{
...
+    path::PathBuf,
+    sync::{Arc, Mutex},
+};
 
 mod sync;
 pub(crate) use sync::{repost_message, RepostAction};
 pub use sync::{AccountSynchronizer, SyncedAccount, TransferMetadata};
 
+type AddressesLock = Arc<Mutex<Vec<IotaAddress>>>;
+type AccountAddressesLock = Arc<Mutex<HashMap<AccountIdentifier, AddressesLock>>>;
```
Alex6323, line 29: sorry, if this is a dumb question, but I am not very much familiar with `wallet.rs` yet. The typedef just looks a bit weird to me.....but are we sure, that `AddressesLock` (the value stored in the `HashMap`) needs to be protected as well as the `HashMap` it is stored in? That would imply that this `Vec<IotaAddress>` needs to be manipulated somewhere else in the codebase. I wonder, though, why that needs to be the case. Couldn't we make those manipulation through the `AccountAddressesLock`.

lucasfernog: I want to lock the transfer process if a input selection is happening. That's because I don't want to have concurrent transfers using the same input addresses; so it needs the wrapper Mutex.

Alex6323: hmmm, okay

PR #174, opened 2020-12-16
refactor(account): keep all accounts in memory

lucasfernog:
# Description of change

Instead of loading the account from storage and saving on each operation, we'll keep the account references stored on the AccountManager struct, wrapped on Arc<RwLock<>> for thread safety. This prevents race conditions between e.g. the polling system and the transfer API, where account data can be lost. 

## Type of change

- Refactor

Unit tests + faucet.

## Change checklist

### DIFF
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -131,6 +131,17 @@ declare_types! {
             Ok(cx.undefined().upcast())
         }
 
+        method loadAccounts(mut cx) {
+            {
+                let this = cx.this();
+                let guard = cx.lock();
+                let ref_ = &this.borrow(&guard).0;
+                let mut manager = ref_.write().unwrap();
+                manager.load_accounts().expect("failed to load accounts");
+            }
+            Ok(cx.undefined().upcast())
+        }
+
         method createAccount(mut cx) {
             let account = {
                 let account_to_create = cx.argument::<JsValue>(0)?;
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -190,8 +190,8 @@ declare_types! {
                 crate::block_on(async move { manager.get_account(&id).await })
             };
             match account {
-                Ok(acc) => {
-                    let id = crate::store_account(acc);
+                Ok(account) => {
+                    let id = crate::store_account(account);
                     let id = cx.string(serde_json::to_string(&id).unwrap());
                     Ok(JsAccount::new(&mut cx, vec![id])?.upcast())
                 },
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -101,8 +101,8 @@ declare_types! {
                 None => Default::default(),
             };
             let manager = match options.storage_type {
-                StorageType::Sqlite => AccountManager::with_storage_adapter(&options.storage_path, SqliteStorageAdapter::new(&options.storage_path, "accounts").unwrap()),
-                StorageType::Stronghold => AccountManager::with_storage_adapter(&options.storage_path, StrongholdStorageAdapter::new(&options.storage_path).unwrap()),
+                StorageType::Sqlite => crate::block_on(AccountManager::with_storage_adapter(&options.storage_path, SqliteStorageAdapter::new(&options.storage_path, "accounts").unwrap())),
+                StorageType::Stronghold => crate::block_on(AccountManager::with_storage_adapter(&options.storage_path, StrongholdStorageAdapter::new(&options.storage_path).unwrap())),
             };
             let manager = manager.expect("error initializing account manager");
             Ok(AccountManagerWrapper(Arc::new(RwLock::new(manager))))
@@ -116,28 +116,6 @@ declare_types! {
                 let ref_ = &this.borrow(&guard).0;
                 let mut manager = ref_.write().unwrap();
                 crate::block_on(async move { manager.set_stronghold_password(password).await }).expect("error setting stronghold password");
-            }
-            Ok(cx.undefined().upcast())
-        }
-
-        method startBackgroundSync(mut cx) {
-            {
-                let this = cx.this();
-                let guard = cx.lock();
-                let ref_ = &this.borrow(&guard).0;
-                let mut manager = ref_.write().unwrap();
-                crate::block_on(async move { manager.start_background_sync().await });
-            }
-            Ok(cx.undefined().upcast())
-        }
-
-        method loadAccounts(mut cx) {
-            {
-                let this = cx.this();
-                let guard = cx.lock();
-                let ref_ = &this.borrow(&guard).0;
-                let mut manager = ref_.write().unwrap();
-                crate::block_on(async move { manager.load_accounts().await }).expect("failed to load accounts");
             }
             Ok(cx.undefined().upcast())
         }
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -137,7 +137,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let mut manager = ref_.write().unwrap();
-                manager.load_accounts().expect("failed to load accounts");
+                crate::block_on(async move { manager.load_accounts().await }).expect("failed to load accounts");
             }
             Ok(cx.undefined().upcast())
         }
@@ -187,7 +187,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let manager = ref_.read().unwrap();
-                manager.get_account(&id)
+                crate::block_on(async move { manager.get_account(&id).await })
             };
             match account {
                 Ok(acc) => {
@@ -224,7 +224,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let manager = ref_.read().unwrap();
-                manager.get_accounts()
+                crate::block_on(async move { manager.get_accounts().await })
             };
 
             let js_array = JsArray::new(&mut cx, accounts.len() as u32);
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -115,7 +115,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let mut manager = ref_.write().unwrap();
-                manager.set_stronghold_password(password).expect("error setting stronghold password");
+                crate::block_on(async move { manager.set_stronghold_password(password).await }).expect("error setting stronghold password");
             }
             Ok(cx.undefined().upcast())
         }
@@ -126,7 +126,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let mut manager = ref_.write().unwrap();
-                manager.start_background_sync();
+                crate::block_on(async move { manager.start_background_sync().await });
             }
             Ok(cx.undefined().upcast())
         }
@@ -170,7 +170,7 @@ declare_types! {
                         .expect("invalid account created at format"),
                     );
                 }
-                builder.initialise().expect("error creating account")
+                crate::block_on(async move { builder.initialise().await }).expect("error creating account")
             };
 
             let id = crate::store_account(account);
@@ -206,11 +206,11 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let manager = ref_.read().unwrap();
-                manager.get_account_by_alias(alias)
+                crate::block_on(async move { manager.get_account_by_alias(alias).await })
             };
             match account {
-                Some(acc) => {
-                    let id = crate::store_account(acc);
+                Some(account) => {
+                    let id = crate::store_account(account);
                     let id = cx.string(serde_json::to_string(&id).unwrap());
                     Ok(JsAccount::new(&mut cx, vec![id])?.upcast())
                 },
@@ -246,7 +246,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let manager = ref_.read().unwrap();
-                manager.remove_account(&id).expect("error removing account")
+                crate::block_on(async move { manager.remove_account(&id).await }).expect("error removing account")
             };
             Ok(cx.undefined().upcast())
         }
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -161,10 +161,11 @@ declare_types! {
                 }
                 builder.initialise().expect("error creating account")
             };
-            let account = serde_json::to_string(&account).unwrap();
-            let account = cx.string(account);
 
-            Ok(JsAccount::new(&mut cx, vec![account])?.upcast())
+            let id = crate::store_account(account);
+            let id = cx.string(serde_json::to_string(&id).unwrap());
+
+            Ok(JsAccount::new(&mut cx, vec![id])?.upcast())
         }
 
         method getAccount(mut cx) {
@@ -179,9 +180,9 @@ declare_types! {
             };
             match account {
                 Ok(acc) => {
-                    let account = serde_json::to_string(&acc).unwrap();
-                    let account = cx.string(account);
-                    Ok(JsAccount::new(&mut cx, vec![account])?.upcast())
+                    let id = crate::store_account(acc);
+                    let id = cx.string(serde_json::to_string(&id).unwrap());
+                    Ok(JsAccount::new(&mut cx, vec![id])?.upcast())
                 },
                 Err(_) => Ok(cx.undefined().upcast())
             }
@@ -198,9 +199,9 @@ declare_types! {
             };
             match account {
                 Some(acc) => {
-                    let account = serde_json::to_string(&acc).unwrap();
-                    let account = cx.string(account);
-                    Ok(JsAccount::new(&mut cx, vec![account])?.upcast())
+                    let id = crate::store_account(acc);
+                    let id = cx.string(serde_json::to_string(&id).unwrap());
+                    Ok(JsAccount::new(&mut cx, vec![id])?.upcast())
                 },
                 None => Ok(cx.undefined().upcast())
             }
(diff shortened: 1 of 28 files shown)


bindings/node/native/src/lib.rs
```diff
@@ -22,102 +21,68 @@ use tokio::runtime::Runtime;
...
 
-pub(crate) fn store_account(account: Account) -> String {
-    let mut map = instances()
+pub(crate) fn store_account(account: AccountGuard) -> AccountIdentifier {
+    let mut map = account_instances()
         .write()
         .expect("failed to lock account instances: store_account()");
-    let id: String = thread_rng().sample_iter(&Alphanumeric).take(10).collect();
-    map.insert(id.clone(), Arc::new(RwLock::new(account)));
+    let id = {
```
zesterer: I'm curious: can this not be done inline? Rust semantics keep any wrapper types (like lock guards) around until the end of an expression to this should be fine to do inline.

src/account/mod.rs
```diff
@@ -242,6 +250,127 @@ pub struct Account {
     has_pending_changes: bool,
 }
 
+/// A thread guard over an account.
+#[derive(Debug, Clone)]
+pub struct AccountGuard(Arc<RwLock<Account>>);
```
zesterer: Guards are specifically named for their ability to provide access to an inner value while upholding some invariant (in this case, mutual exclusion of writers). A better name for this type would be `SharedAccount`, `AccountHandle`, `AccountRef`, etc.

src/account/mod.rs
```diff
@@ -242,6 +250,127 @@ pub struct Account {
...
+        account.has_pending_changes = true;
+
+        let _ = crate::monitor::monitor_address_balance(self.clone(), address.address());
+
+        Ok(address)
+    }
+
+    /// Bridge to [Account#latest_address](struct.Account.html#method.latest_address).
+    pub fn latest_address(&self) -> Option<Address> {
+        let account = self.0.read().unwrap();
```
zesterer: I don't think there should be any problem with performing these operations inline (i.e: without the intermediate `account` binding).

src/account/mod.rs
lucasfernog: this makes sense, i'll change it

bindings/node/native/src/lib.rs
lucasfernog: oh you're right, I guess there's a lot of places I can use it. I thought I needed to save the reference so it doesn't go out of scope.

src/account/mod.rs
lucasfernog: yeah thought about it when you mentioned on the other comment :) thanks for the tip

src/account/mod.rs
```diff
@@ -242,6 +250,127 @@ pub struct Account {
...
+        AccountSynchronizer::new(self.clone())
+    }
+
+    /// Gets a new unused address and links it to this account.
+    pub fn generate_address(&self) -> crate::Result<Address> {
+        let mut account = self.0.write().unwrap();
+        let address = crate::address::get_new_address(&account)?;
+        account.addresses.push(address.clone());
+
+        account.has_pending_changes = true;
```
zesterer: This looks like it's a flag that informs some other system of an update. If this code is going to be expanded later with more operations that might need to set this flag, I recommend making the fields of `Account` private, creating immutable accessor methods for them, and creating a dedicated method on `Account` to automatically do this when mutating. Something like:

```rust
impl Account {
    pub fn do_mut<R>(&mut self, f: impl FnOnce(&mut Self) -> R) -> R {
        let res = f(self);
        self.has_pending_changes = true;
        res
    }
}

// Usage
self.0.write().unwrap().do_mut(|account| {
    let addr = crate::address::get_new_address(account)?;
    account.addresses.push(addrclone());
    Ok(addr)
})
```

This will prevent accidental incorrect usage of the API which might result in a very difficult to debug error later on.

src/actor/mod.rs
```diff
@@ -344,14 +367,24 @@ mod tests {
         let tx = spawn_actor();
 
         // create an account
-        let account = AccountToCreate::default();
+        let account = AccountToCreate {
+            client_options: ClientOptionsBuilder::node("http://node.iota").unwrap().build(),
+            mnemonic: None,
+            alias: None,
+            created_at: None,
+        };
```
zesterer, line 382: If this type has a default impl with `None`s, you can just do

```rust
let account = AccountToCreate {
    client_options: ClientOptionsBuilder::node("http://node.iota").unwrap().build(),
    ..Default::default()
};
```

src/monitor.rs
```diff
@@ -143,94 +126,88 @@ async fn process_output(
...
     let message_id_ = *message_id;
-    mutate_account(&account_id, &storage_path, |account| {
-        {
-            let addresses = account.addresses_mut();
-            let address_to_update = addresses.iter_mut().find(|a| a.address() == &address).unwrap();
-            address_to_update.handle_new_output(address_output);
-            crate::event::emit_balance_change(&account_id, &address_to_update, *address_to_update.balance());
-        }
+    {
+        let addresses = account.addresses_mut();
```
zesterer: Single use of binding?

src/monitor.rs
```diff
@@ -143,94 +126,88 @@ async fn process_output(
...
-                crate::event::emit_transaction_event(
-                    crate::event::TransactionEventType::NewTransaction,
-                    &account_id,
-                    &message,
-                );
-                account.messages_mut().push(message);
-            }
+    match account.messages_mut().iter().position(|m| m.id() == &message_id_) {
+        Some(message_index) => {
+            let message = &mut account.messages_mut()[message_index];
```
zesterer: Single use of binding?

bindings/node/native/src/classes/account_manager/mod.rs
```diff
@@ -179,9 +191,9 @@ declare_types! {
             };
             match account {
                 Ok(acc) => {
```
Alex6323: hard to find anything to criticize, especially when not too familiar with the code base, but my preference here would be to just use the same name (`account`) on the inner variable again. Specially this abbreviation is also used to abbreviate `accumulator`.

src/monitor.rs
```diff
@@ -143,94 +126,88 @@ async fn process_output(
...
-                    let message = messages.iter_mut().find(|m| m.id() == &message_id).unwrap();
-                    message.set_confirmed(Some(confirmed));
-                    *message.id()
-                };
-
-                crate::event::emit_confirmation_state_change(&account_id, &message, confirmed);
-            })?;
+            let mut account = account.write().unwrap();
+            {
+                let messages = account.messages_mut();
```
zesterer, line 207: Single use of binding

src/actor/mod.rs
```diff
@@ -344,14 +367,24 @@ mod tests {
...
         send_message(&tx, MessageType::SetStrongholdPassword("password".to_string())).await;
         let response = send_message(&tx, MessageType::CreateAccount(account)).await;
         match response.response() {
             ResponseType::CreatedAccount(created_account) => {
-                // remove the created account
-                let response = send_message(&tx, MessageType::RemoveAccount(created_account.id().clone())).await;
-                assert!(matches!(response.response(), ResponseType::RemovedAccount(_)));
+                let id = created_account.id().clone();
+                std::thread::spawn(move || {
```
zesterer, line 388: Why spawn a thread here? If there's an async executor running it would be better to just create an async task

bindings/node/native/src/classes/synced_account/mod.rs
```diff
@@ -17,15 +13,19 @@ mod repost;
 mod send;
 
 #[derive(Clone)]
-pub struct SyncedAccountWrapper(Arc<RwLock<SyncedAccount>>, String);
+pub struct SyncedAccountWrapper(pub String);
```
Alex6323, line 16: again, more a matter of taste than an issue. This struct is `pub` so I guess it gets exported and is part of the API of the library. In that case I would name that type `SyncedAccount`, and the type that it wraps `SyncedAccountInner`. That makes the API nicer to read IMHO.

src/account_manager.rs
```diff
@@ -76,17 +82,35 @@ impl AccountManager {
...
+            accounts: Default::default(),
+            stop_polling: Arc::new(AtomicBool::new(false)),
+            polling_handle: None,
         };
         Ok(instance)
     }
 
+    /// Loads the account from the storage into the manager instance.
+    pub fn load_accounts(&mut self) -> crate::Result<()> {
+        let empty_accounts = {
```
zesterer: This could be compute inline in the `if` predicate

src/account_manager.rs
```diff
@@ -95,11 +119,28 @@ impl AccountManager {
...
             let monitoring_disabled = self.start_monitoring().is_err();
-            self.start_polling(monitoring_disabled);
+            self.polling_handle = Some(self.start_polling(monitoring_disabled));
             self.started_monitoring = true;
         }
     }
 
+    /// Stops the background polling and MQTT monitoring.
+    pub fn stop_background_sync(&mut self) -> crate::Result<()> {
+        {
```
zesterer: If done inline, the outer block wouldn't be needed here

src/account_manager.rs
```diff
@@ -95,11 +119,28 @@ impl AccountManager {
...
+            let accounts = self.accounts.read().unwrap();
+            for account in accounts.values() {
+                let _ = crate::monitor::unsubscribe(account.clone());
+            }
+        }
+
+        if let Some(handle) = self.polling_handle.take() {
+            self.stop_polling.store(true, Ordering::Relaxed);
```
zesterer: The polling thread could instead become an async task. Making this change would be better since it might then `select!` over a oneshot channel so that it shuts down immediately instead of when the next polling period comes around. This would both reduce CPU usage (no need for frequent wakeups) and reduce shutdown time (no need to wait for the next wakeup).

src/account_manager.rs
```diff
@@ -137,32 +184,54 @@ impl AccountManager {
                         // when the error is dropped, the on_error event will be triggered
                     }
                 });
-                thread::sleep(interval);
+
+                thread::sleep(sleep_duration);
```
zesterer: Why are there two sleeps here?

src/account_manager.rs
```diff
@@ -171,22 +240,24 @@ impl AccountManager {
...
-            .ok_or_else(|| anyhow::anyhow!("destination account address list empty"))?
-            .clone();
-        let from_synchronized = from_account.sync().execute().await?;
-        let metadata = from_synchronized
+    ) -> crate::Result<Message> {
+        let from_account_guard = self.get_account(from_account_id)?;
+        let to_account_guard = self.get_account(to_account_id)?;
+
+        let to_address = {
+            let to_account = to_account_guard.read().unwrap();
```
zesterer: Doing this inline would be better, no outer block needed

src/account_manager.rs
```diff
@@ -171,22 +240,24 @@ impl AccountManager {
...
+            to_account
+                .latest_address()
+                .ok_or_else(|| anyhow::anyhow!("destination account address list empty"))?
```
zesterer: Out of interest, is it necessary to use `anyhow`? The crate dynamically allocates space for errors. It's often more performant or even easier to just use a enum. This also means that any crates built on top of this API can handle different errors in different ways instead of needing to inspect an opaque error type. `anyhow` should really only be used for user-facing applications where no program needs to consume the error higher up the stack (and even in such cases I'd argue it's not necessarily a good fit).

src/account_manager.rs
```diff
@@ -259,62 +330,65 @@ impl AccountManager {
...
     /// Gets the account associated with the given alias (case insensitive).
-    pub fn get_account_by_alias<S: Into<String>>(&self, alias: S) -> Option<Account> {
+    pub fn get_account_by_alias<S: Into<String>>(&self, alias: S) -> Option<AccountGuard> {
         let alias = alias.into().to_lowercase();
-        if let Ok(accounts) = self.get_accounts() {
-            accounts.into_iter().find(|acc| acc.alias().to_lowercase() == alias)
-        } else {
-            None
-        }
+        self.get_accounts().into_iter().find(|acc| {
```
zesterer: This method invokes `get_accounts` which takes the entire account `HashMap`, iterates through the elements, clones them, collects them into a vector, and then searches that vector for the required element by cloning the alias string into a new allocation, converting it to lowercase, and then comparing it to the desired string. This comes with a very significant amount of unnecessary overhead. If you must use linear search (an alias -> account mapping would be better) then I suggest something like this instead:

```rust
self.accounts.read().unwrap().values().find(|a| {
    // alias() should return a &str rather than a new string
    a.alias().chars().zip(alias.chars()).all(|(x, y)| x.to_lowercase() == y)
})
```

The superior method to this would be to maintain a mapping of the lowercase account name to the account itself such that the only thing required is an `O(1)` hashmap lookup.

bindings/node/native/src/lib.rs
```diff
@@ -22,102 +21,68 @@ use tokio::runtime::Runtime;
...
+
+/// Gets the synced account instances map.
+fn synced_account_instances() -> &'static SyncedAccountInstanceMap {
+    static INSTANCES: Lazy<SyncedAccountInstanceMap> = Lazy::new(Default::default);
+    &INSTANCES
 }
 
-pub(crate) fn remove_account(id: &str) {
-    let mut map = instances()
+pub(crate) fn get_synced_account(id: &str) -> SyncedAccountGuard {
```
Alex6323: probably already on the todo list: error handling and return a result

src/account/mod.rs
```diff
@@ -17,13 +18,14 @@ use serde::{Deserialize, Serialize};
...
+    sync::{Arc, Mutex, RwLock},
 };
 
 mod sync;
 pub(crate) use sync::{repost_message, RepostAction};
-pub use sync::{AccountSynchronizer, SyncedAccount, TransferMetadata};
+pub use sync::{AccountSynchronizer, SyncedAccount};
 
 type AddressesLock = Arc<Mutex<Vec<IotaAddress>>>;
 type AccountAddressesLock = Arc<Mutex<HashMap<AccountIdentifier, AddressesLock>>>;
```
Alex6323, line 32: my preference would be to name it just `Addresses` and `AccountAddresses` if there aren't already types named that way. I don't see much value in signaling that you first need to get a lock before being able to change the shared data. Ignore if you disagree :smile:

lucasfernog: I'll refactor this on a subsequent PR.

src/account_manager.rs
lucasfernog: yeah it's on my todo list and I think i'll do it today

bindings/node/native/src/lib.rs
lucasfernog: in this case it's ok because a panic shouldn't happen here

bindings/node/native/src/classes/synced_account/mod.rs
lucasfernog: this is just a hidden type on the JavaScript class that the binding generates for the synced account so I think it's fine to leave it as is; it's only public because the JS-generated code is also public.

src/actor/mod.rs
lucasfernog: this is just a test case that will be changed soon

src/actor/mod.rs
lucasfernog: yeah but in this case `client_options` is required so i'll keep it like that
although this gives me an idea to deserialize from the AccountInitialiser instead, just like bee's logger
I'll do this later

src/account/mod.rs
lucasfernog: amazing idea <3

lucasfernog:
Next steps (separate PRs):

- remove anyhow usage
- add AccountManager builder
- stronghold refactor

bindings/node/native/src/lib.rs
```diff
@@ -22,102 +21,61 @@ use tokio::runtime::Runtime;
...
-pub(crate) fn update_account(id: &str, account: Account) {
-    let mut map = instances()
+/// Gets the synced account instances map.
+fn synced_account_instances() -> &'static SyncedAccountInstanceMap {
+    static INSTANCES: Lazy<SyncedAccountInstanceMap> = Lazy::new(Default::default);
+    &INSTANCES
+}
+
+pub(crate) fn get_synced_account(id: &str) -> SyncedAccountHandle {
+    let map = synced_account_instances()
```
zesterer: This can probably be done inline

src/account_manager.rs
```diff
@@ -41,67 +52,89 @@ pub struct AccountManager {
...
+            let _ = crate::block_on(async {
+                for account in accounts.read().await.values() {
+                    let _ = crate::monitor::unsubscribe(account.clone());
+                }
+
+                if let Some(sender) = stop_polling_sender {
+                    sender.send(()).expect("failed to stop polling process");
+                }
+            });
+        });
```
zesterer: Just so you're aware, spawning a thread and not waiting for it to be joined opens up the possibility of the thread being aborted on program exit. I don't understand the ramifications of this in the context of your code, so this might not be an issue at all: but if the thread should run to completion then its `JoinHandle` needs to be given to something that can wait for it to finish.

src/account_manager.rs
```diff
@@ -110,59 +143,95 @@ impl AccountManager {
...
+                    loop {
+                        tokio::select! {
+                            _ = async {
+                                let storage_path_ = storage_path.clone();
+                                let accounts_ = accounts.clone();
+
+                                if let Err(panic) = AssertUnwindSafe(poll(accounts_.clone(), storage_path_, is_monitoring_disabled))
+                                    .catch_unwind()
+                                    .await {
+                                    let msg = if let Some(message) = panic.downcast_ref::<String>() {
```
zesterer: Using `Cow<str>` to allow for both `String` and `str` without needing dynamic dispatching and downcasting might be a good idea (the code seems to imply that the panic message being neither a `&'static str` nor a `String` is a type system hole: `Cow` will let you patch it)

Edit: I see `std`'s API requires the dynamic dispatch. That's a shame, but I still think that using `Cow` is warranted here. On the case where the downcast fails, I think the program should simply panic. If this fails then there's something more serious going on and it's probably safer for the program to abort in such a case. It also has the advantage of making it easier for you to find future bugs rather than wondering where a "system error" string is coming from.

src/account_manager.rs
```diff
@@ -244,12 +311,14 @@ impl AccountManager {
...
-                    Some("password"),
-                )
+                stronghold
+                    .account_import(
+                        Some(created_at_timestamp),
+                        Some(created_at_timestamp),
+                        stronghold_account.mnemonic().to_string(),
+                        Some("password"),
+                    )
+                    .map_err(|e| e.into())
```
zesterer: FYI you can replace single-method closures with the function directly (also means the compiler spends less time doing type inference) like `.map_err(Into::into)`. This also works with enum variants like `Ok`/`Some`/etc. because the compiler considers them to be functions rather than just type constructors.

src/account_manager.rs
```diff
@@ -259,62 +328,74 @@ impl AccountManager {
...
+        let synced_accounts = sync_accounts(accounts.clone(), &storage_path, Some(0)).await?;
+        let accounts_after_sync = accounts.read().await;
 
         // compare accounts to check for balance changes and new messages
         for account_before_sync in &accounts_before_sync {
             let account_after_sync = accounts_after_sync
-                .iter_mut()
-                .find(|account| account.id() == account_before_sync.id())
-                .unwrap();
+                .iter()
```
zesterer: Be aware that this is an O(n^2) operation. If it can be replaced with a hashmap lookup that would likely result in wins on workloads with many accounts.

lucasfernog: nice catch, totally forgot to change it since it`s now a HashMap instead of a Vec

src/account_manager.rs
lucasfernog: I'll check again if this code really needs a catch_unwind; I think with this refactor it won't be needed anymore.

But I do have this check on the Node.js binding and unfortunately I can't remove it there, so I guess i'll write something to improve it.

zesterer (approved):

rajivshah3 (approved):

PR #176, opened 2020-12-18
feat(errors): rename to `Error`, remove anyhow calls

lucasfernog:
# Description of change

Renames the lib's error struct to `Error` and removes anyhow calls, adding explicit error kinds for each possible scenario. 

## Type of change

## Change checklist

### DIFF
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -153,9 +153,9 @@ impl AccountInitialiser {
         let mut accounts = self.accounts.write().await;
 
         let alias = self.alias.unwrap_or_else(|| format!("Account {}", accounts.len()));
-        let signer_type = self
-            .signer_type
-            .ok_or_else(|| anyhow::anyhow!("account signer type is required"))?;
+        let signer_type = self.signer_type.ok_or(crate::Error::AccountInitialiseRequiredField(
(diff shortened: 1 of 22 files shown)


src/account/mod.rs
```diff
@@ -207,13 +207,11 @@ impl AccountInitialiser {
 
 pub(crate) fn account_id_to_stronghold_record_id(account_id: &AccountIdentifier) -> crate::Result<[u8; 32]> {
     if let AccountIdentifier::Id(id) = account_id {
-        let decoded = hex::decode(id).map_err(|_| anyhow::anyhow!("account id must be a hex string"))?;
-        let id: [u8; 32] = decoded
-            .try_into()
-            .map_err(|_| anyhow::anyhow!("invalid account id length"))?;
+        let decoded = hex::decode(id)?;
+        let id: [u8; 32] = decoded.try_into().unwrap();
```
rajivshah3, line 211: If this panics, won't the error be `id can't be index` instead of `invalid account id length`?

rajivshah3 (approved):

src/account/mod.rs
lucasfernog: yeah you're right but I'll change this with the stronghold refactor so I just left it like this for now

PR #179, opened 2020-12-18
refactor(storage): async trait

lucasfernog:
# Description of change

Refactor the storage adapter so all the functions are now async.
This will be needed with the upcoming stronghold update. 

## Type of change

Faucet.

## Change checklist

### DIFF
--- a/bindings/node/native/src/classes/account_manager/mod.rs
+++ b/bindings/node/native/src/classes/account_manager/mod.rs
@@ -298,7 +298,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let manager = ref_.read().unwrap();
-                manager.import_accounts(source).expect("error importing accounts");
+                crate::block_on(manager.import_accounts(source)).expect("error importing accounts");
             };
             Ok(cx.undefined().upcast())
         }
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -22,6 +22,7 @@ use std::{
     ops::Deref,
     path::PathBuf,
     sync::{Arc, Mutex},
+    thread,
 };
 
 mod sync;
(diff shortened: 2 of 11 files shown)


rajivshah3 (approved):

bindings/node/native/src/classes/account_manager/mod.rs
```diff
@@ -298,7 +298,7 @@ declare_types! {
                 let guard = cx.lock();
                 let ref_ = &this.borrow(&guard).0;
                 let manager = ref_.read().unwrap();
-                manager.import_accounts(source).expect("error importing accounts");
+                crate::block_on(manager.import_accounts(source)).expect("error importing accounts");
```
thibault-martinez, line 301: Are we sure we want to block here ?

src/account/mod.rs
```diff
@@ -372,6 +373,21 @@ impl AccountHandle {
     }
 }
 
+impl Drop for Account {
+    fn drop(&mut self) {
+        if self.has_pending_changes {
+            let storage_path = self.storage_path.clone();
+            let account_id = self.id().clone();
+            if let Ok(data) = serde_json::to_string(&self) {
+                let _ = thread::spawn(move || {
```
thibault-martinez, line 382: Why not spawn a task instead of a full thread ?

src/account_manager.rs
```diff
@@ -677,7 +682,8 @@ mod tests {
                 .expect("invalid node URL")
                 .build();
 
-            crate::block_on(async move {
+            let mut runtime = tokio::runtime::Runtime::new().unwrap();
+            runtime.block_on(async move {
```
thibault-martinez, line 686: This looks like a one time block so I assume it's fine

lucasfernog: this is just a test case; it will be a regular async test once I get the new stronghold up and running

bindings/node/native/src/classes/account_manager/mod.rs
lucasfernog: yeah this is fine for now but I'm planning on making this an async API in the future

src/account_manager.rs
thibault-martinez: Whoops, I overlooked it, sorry

src/account/mod.rs
lucasfernog: I need the thread here to use my tokio runtime :(

PR #180, opened 2020-12-18
fix(env_mnemonic): dont prompt for save if the given mnemonic is on env

lucasfernog:
# Description of change

If the user passes the mnemonic that's already saved on the env to the account initialiser, ignore the prompt with info since it's already managed by the user. 

## Type of change

The faucet uses the wallet library this way so it's a way to test it (it always passes the mnemonic to the account creation and it already saves it to it's own .env)

## Change checklist


nothingismagick:
Might this be an attack vector if the attacker can set the ENV?

nothingismagick:
Maybe better to also hide this behind a buildtime feature?

lucasfernog:
yeah tbh this feature is only meant to be provisory to get the testnet up and running; I think we should rethink this later or maybe force stronghold usage.

cvarley100 (approved):

cvarley100:
@lucasfernog - maybe add a `TODO` and a warning

PR #195, opened 2021-01-11
fix(message): indexation's `data` field is optional

lucasfernog:
# Description of change

Makes the `data` field on the indexation optional. 

## Type of change

## Change checklist

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -33,6 +33,32 @@ impl Default for RemainderValueStrategy {
     }
 }
 
+/// The message's index builder.
+#[derive(Debug, Clone, Deserialize)]
+pub struct IndexationBuilder {
+    index: String,
+    data: Option<Vec<u8>>,
+}
+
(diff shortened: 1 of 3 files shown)


src/message.rs
```diff
@@ -65,7 +91,7 @@ impl TransferBuilder {
     }
 
     /// (Optional) message indexation.
-    pub fn with_indexation(mut self, indexation: Indexation) -> Self {
+    pub fn with_indexation(mut self, indexation: IndexationBuilder) -> Self {
```
rajivshah3: Should the method name be changed to `with_indexation_builder`?

laumair (approved):

src/message.rs
lucasfernog: yeah this is looking weird both ways, so I'm just going to rewrite it to have the TransferBuilder type with the original Indexation struct, and deserialize with a wrapper struct.

PR #217, opened 2021-01-14
chore(lib): remove iota::Address::Wots reference, add Payload suffix to message payload types

lucasfernog:


rajivshah3 (approved):
May WOTS rest in peace

PR #254, opened 2021-01-25
Docs/setup ci

huhn511:
Deploys mdbook and rust docs to github pages with a github action.


.github/workflows/deploy_docs_to_gh-pages.yml
```diff
@@ -0,0 +1,24 @@
+name: github pages
+
+on:
+  push:
+    branches:
+      - develop
+jobs:
+  deploy:
+    runs-on: ubuntu-18.04
```
rajivshah3:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/deploy_docs_to_gh-pages.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  push:
+    branches:
+      - develop
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup mdBook
```
rajivshah3, line 20:
```suggestion
      - name: Install rust-docs with nightly toolchain
        uses: actions-rs/toolchain@v1
        with:
            profile: minimal
            toolchain: nightly
            override: true
            components: rust-docs
          
      - name: Setup mdBook
```

I think this needs to be run because the default toolchain is `stable`

rajivshah3 (approved):

PR #297, opened 2021-02-05
fix(polling): deadlock on retry

lucasfernog:
# Description of change

Fixes a deadlock on the automatic retry feature of the polling system. Also changes it so messages that doesn't need to promote/reattach will get their status updated.

N/A

## Type of change

Faucet.

## Change checklist


src/account_manager.rs
```diff
@@ -998,44 +999,69 @@ async fn poll(accounts: AccountStore, storage_file_path: PathBuf, is_mqtt_monito
...
+        for message in &retried_data.reattached {
+            emit_transaction_event(TransactionEventType::Reattachment, &account_id, &message);
+        }
+
+        account.append_messages(retried_data.reattached);
+        account.append_messages(retried_data.promoted);
+
+        for message_id in retried_data.no_need_promote_or_reattach {
+            let message = account.get_message_mut(&message_id).unwrap();
+            if let Ok(metadata) = client.read().await.get_message().metadata(&message_id).await {
```
Thoralf-M, line 1072: You shouldn't assume that it's confirmed here, if you send a message and call get metadata it could return that reattachment and promotion isn't needed and then still get left behind

lucasfernog: in this case, isn't the message unconfirmed?

Thoralf-M: Yes it would be unconfirmed
Sry I thought you would always just set it confirmed, but you compare it

Thoralf-M (approved):

PR #324, opened 2021-02-17
fix(transfer): allow to input address if outputs.len() > 1

lucasfernog:
# Description of change

Allow transfer to the input address if it has more than one output (the use case is a manual output consolidation).

N/A

## Type of change

## Change checklist


rajivshah3 (approved):

Thoralf-M:
Why don't just allow to send to the same address, even if there is only one output? I know that's useless in the sense that it doesn't really change anything at the end, but if someones wants to try sending transactions but don't want to send the iotas to someone else he could think about just sending it to his own address and then he needs to create a second address to get it working

lucasfernog:
I've discussed this with @laumair and we agreed that this is pointless and we shouldn't allow it.

PR #333, opened 2021-02-18
fix(bindings/nodejs): Run scripts with NPM instead of Yarn

rajivshah3:
# Description of change

NPM users who don't have Yarn installed are not able to use the JS bindings since they explicitly use Yarn in the `install` and `rebuild` scripts. ~~TIL we can use `$npm_execpath` to call either Yarn or NPM (https://stackoverflow.com/a/55440486/3188334).~~ This PR uses NPM instead of Yarn since NPM is included with Node.js

N/A

## Type of change

Tested with Yarn and NPM

## Change checklist

### DIFF
--- a/.changes/js-bindings-npm-scripts.md
+++ b/.changes/js-bindings-npm-scripts.md
@@ -1,0 +1,5 @@
+---
+"nodejs-binding": patch
+---
+
+Fixes issues with the installation script when using with NPM instead of Yarn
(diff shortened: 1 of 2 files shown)


rajivshah3:
Windows doesn't support using `$npm_execpath`, so we can just use NPM instead. We can reasonably assume NPM is installed since it's included with Node.js

lucasfernog (approved):

PR #349, opened 2021-02-25
fix(stronghold): map referenced signatures with both index and internal

lucasfernog:
# Description of change

This PR fixes an issue when two addresses of the same index are part of a transaction (a public address and a change address), the signature would be invalid since the second signature would be considered a reference of the first one.

N/A

## Type of change

## Change checklist

### DIFF
--- a/src/signing/stronghold.rs
+++ b/src/signing/stronghold.rs
@@ -62,7 +62,7 @@ impl super::Signer for StrongholdSigner {
         inputs.sort_by(|a, b| a.input.cmp(&b.input));
 
         for (current_block_index, recorder) in inputs.iter().enumerate() {
-            let signature_index = format!("{}/{}", recorder.address_index, recorder.address_internal);
+            let signature_index = format!("{}{}", recorder.address_index, recorder.address_internal);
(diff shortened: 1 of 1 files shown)


src/signing/stronghold.rs
```diff
@@ -58,13 +58,14 @@ impl super::Signer for StrongholdSigner {
         let hashed_essence = essence.hash();
 
         let mut unlock_blocks = vec![];
-        let mut signature_indexes = HashMap::<usize, usize>::new();
+        let mut signature_indexes = HashMap::<String, usize>::new();
         inputs.sort_by(|a, b| a.input.cmp(&b.input));
 
         for (current_block_index, recorder) in inputs.iter().enumerate() {
+            let signature_index = format!("{}/{}", recorder.address_index, recorder.address_internal);
```
Thoralf-M: / separator is not needed

Thoralf-M (approved):

src/signing/stronghold.rs
lucasfernog: thx

PR #361, opened 2021-03-02
refactor(message): add input details on transaction essence inputs

lucasfernog:
# Description of change

Fetch input information from the node /output/{output_id} endpoint and store it on the Message object.

N/A

## Type of change

CLI wallet.

## Change checklist

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/bindings/python/native/src/types/address.rs
+++ b/bindings/python/native/src/types/address.rs
@@ -21,12 +21,12 @@ pub struct Address {
     outputs: Vec<AddressOutput>,
 }
 
-#[derive(Debug, DeriveFromPyObject, DeriveIntoPyObject)]
+#[derive(Debug, Clone, DeriveFromPyObject, DeriveIntoPyObject)]
 pub struct AddressWrapper {
     inner: String,
 }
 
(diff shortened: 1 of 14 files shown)


bindings/python/native/src/types/address.rs
```diff
@@ -105,7 +105,11 @@ impl From<RustWalletAddress> for Address {
             balance: *wallet_address.balance(),
             key_index: *wallet_address.key_index(),
             internal: *wallet_address.internal(),
-            outputs: wallet_address.outputs().iter().map(|output| output.into()).collect(),
+            outputs: wallet_address
+                .outputs()
+                .iter()
+                .map(|output| output.clone().into())
```
Thoralf-M: Doesn't it work without the clone?

Thoralf-M (approved):

bindings/python/native/src/types/address.rs
lucasfernog: it can work if a duplicate the `From` impl :( i'll do it

lucasfernog: actually i'll just change the other usage instead

PR #422, opened 2021-03-13
Make input selection aware of dust

Thoralf-M:
# Description of change

Check if selected inputs would leave dust behind and select more if that would be the case.
Create MAX_INPUT_SELECTION_TRIES so branch_and_bound doesn't take "forever" with many inputs

## Type of change

Tests

## Change checklist

### DIFF
--- a/src/account/sync/input_selection.rs
+++ b/src/account/sync/input_selection.rs
@@ -3,7 +3,13 @@
 
 use crate::address::AddressWrapper;
 use rand::{prelude::SliceRandom, thread_rng};
-use std::{cmp::Ordering, convert::TryInto};
+use std::{
+    cmp::Ordering,
+    sync::atomic::{AtomicI64, Ordering as AtomicOrdering},
+};
+
+const DUST_ALLOWANCE_VALUE: u64 = 1_000_000;
+const MAX_INPUT_SELECTION_TRIES: i64 = 10_000_000;
 
 #[derive(Debug, Clone)]
 pub struct Input {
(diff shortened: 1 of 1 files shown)


src/account/sync/input_selection.rs
```diff
@@ -13,23 +19,44 @@ pub struct Input {
...
 pub fn select_input(target: u64, mut available_utxos: Vec<Input>) -> crate::Result<Vec<Input>> {
-    if target > available_utxos.iter().fold(0, |acc, address| acc + address.balance) {
+    let total_available_balance = available_utxos.iter().fold(0, |acc, address| acc + address.balance);
+    if target > total_available_balance {
         return Err(crate::Error::InsufficientFunds);
     }
 
+    // Not insufficient funds, but still not possible to create this transaction because it would create dust
+    if target != total_available_balance
+        && target as i64 > (total_available_balance as i64 - DUST_ALLOWANCE_VALUE as i64)
```
lucasfernog: shouldn't the condition here be `total_available_balance - target < DUST_ALLOWANCE_VALUE`? also this doesn't check if the address has a dust allowance output :(

lucasfernog: also using `i64` here might lead to a panic if the value doesn't fit.

lucasfernog (changes requested):

src/account/sync/input_selection.rs
Thoralf-M: A valid iota amount would always fit since the whole supply is 2779530283277761 and i64::MAX 9223372036854775807

Regarding the condition, it checks the same I think, since we check in the previous `target > total_available_balance`, but I can change it so it's clearer

lucasfernog: oh you're right, i just understood your condition lol

lucasfernog (approved):

PR #424, opened 2021-03-13
Change input selection

Thoralf-M:
# Description of change

Changes the input selection to work based on single outputs and not on address basis, because that could make it impossible to move funds if addresses have many outputs, also changes the input selection so transactions from one address to the same address again are allowed when the balance of it is <= the transfer amount, because in that case we need at least on other input address which we can use as remainder if the transfer amount is > balance of the first input address

## Type of change

## Change checklist


src/account/sync/mod.rs
```diff
@@ -952,7 +929,15 @@ impl SyncedAccount {
         );
 
         let remainder = if has_remainder {
-            selected_addresses.last().cloned()
+            selected_addresses
+                .clone()
```
lucasfernog, line 1142: oh i just realized you're cloning the entire list here; this can be improved by cloning the appropriated item instead.

laumair:
What is the significance of allowing this if available balance is less than transfer value? To me it makes sense, to allow this if there is an exact match i.e., available balance is equal to transfer object. Allowing address with less available balance can still create remainders (unless I am missing something) ?

Thoralf-M:
@laumair if the balance is less than the transfer value you need at least one other address which can then be used as remainder address

src/account/sync/mod.rs
```diff
@@ -1099,70 +1104,116 @@ impl SyncedAccount {
...
         let remainder = if has_remainder {
-            selected_addresses.last().cloned()
+            let input_for_remainder = selected_outputs
+                .iter()
+                // We filter the output addresses, but since we checked that this address balance <=
+                // transfer_obj.amount.get() we need to have another input address
+                .filter(|input| !transfer_obj.outputs.iter().any(|o| o.address == input.output.address))
+                .collect::<Vec<&input_selection::Input>>()
+                .last()
+                .cloned()
```
melatron, line 1179: Do we need the second cloned()?

src/account/sync/mod.rs
```diff
@@ -1259,62 +1310,55 @@ impl SyncedAccount {
...
-                        )],
+                        vec![input_selection::AddressInputs {
+                            internal: *address.internal(),
+                            address: address.address().clone(),
+                            outputs: address_inputs,
+                        }],
                         None,
                     )
                 } else {
                     // TODO
```
melatron: What does this TODO stands for?

src/account/sync/input_selection.rs
```diff
@@ -41,36 +68,59 @@ pub fn select_input(target: u64, mut available_utxos: Vec<Input>) -> crate::Resu
...
         Ok(selected_coins)
     } else {
         // If no match, Single Random Draw
-        Ok(single_random_draw(target, available_utxos))
+        let mut signature_locked_outputs_ = signature_locked_outputs.clone();
+        let mut dust_allowance_outputs_ = dust_allowance_outputs.clone();
+        signature_locked_outputs_.shuffle(&mut thread_rng());
+        dust_allowance_outputs_.shuffle(&mut thread_rng());
+        let mut inputs = single_draw(target, signature_locked_outputs_, dust_allowance_outputs_);
+        if inputs.len() > max_inputs {
```
melatron: Don't we need to sort the dust_allowance_outputs?

src/account/sync/input_selection.rs
```diff
@@ -156,9 +206,8 @@ mod tests {
                 .build()
                 .unwrap();
             available_utxos.push(super::Input {
-                address: address.address().clone(),
-                balance: address.balance(),
                 internal: false,
+                output: address.outputs().iter().next().expect("No output").1.clone(),
```
melatron: Maybe it would be nice to have a comment for this line.

src/account/sync/input_selection.rs
```diff
@@ -179,9 +228,9 @@ mod tests {
...
             let mut available_utxos = generate_random_utxos(&mut rng, 25);
             let sum_utxos_picked = sum_random_utxos(&mut rng, &mut available_utxos);
-            let selected = select_input(sum_utxos_picked, available_utxos).unwrap();
+            let selected = select_input(sum_utxos_picked, available_utxos, 127).unwrap();
             assert_eq!(
-                selected.iter().fold(0, |acc, address| { acc + address.balance }),
+                selected.iter().fold(0, |acc, input| { acc + input.output.amount }),
                 sum_utxos_picked
             );
         }
```
melatron, line 235: Wouldn't global const values for 127, 25, 5 and so on be better?

src/account/sync/mod.rs
Thoralf-M: Not sure, maybe there should be a different error, because the input address was just not found

src/account/sync/mod.rs
Thoralf-M: Yes, because last() returns `Option<&T>`

src/account/sync/input_selection.rs
Thoralf-M: That's just for the tests, I don't think we should have global consts for them

src/account/sync/input_selection.rs
Thoralf-M: Will refactor the whole section

src/account/sync/input_selection.rs
Thoralf-M: Yes

melatron (approved):

PR #439, opened 2021-03-16
feat(sync): emit all events at once when syncing all accounts

lucasfernog:


src/account/sync/mod.rs
```diff
@@ -627,7 +640,8 @@ impl AccountSynchronizer {
         self
     }
 
-    /// Do not emit events. Useful on account discovery.
+    /// Do not emit events and return then on the SyncedAccount object instead.
```
Thoralf-M: Should it be `them`?

Thoralf-M (approved):

PR #451, opened 2021-03-20
fix(mqtt): new address monitoring

lucasfernog:
# Description of change

Properly `.await` on mqtt async tasks.

N/A

## Type of change

CLI wallet, getting funds on new address and checking if MQTT is subscribed/responding.

### DIFF
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -522,9 +522,17 @@ impl AccountHandle {
             })
             .await?;
 
-        let _ = crate::monitor::monitor_address_balance(self.clone(), address.address());
+        // monitor on a non-async function to prevent cycle computing the `monitor_address_balance` fn type
+        self.monitor_address(address.address().clone());
 
         Ok(address)
+    }
+
+    pub(crate) fn monitor_address(&self, address: AddressWrapper) {
+        let handle = self.clone();
+        crate::spawn(async move {
+            let _ = crate::monitor::monitor_address_balance(handle, &address).await;
+        });
     }
 
     /// Synchronizes the account addresses with the Tangle and returns the latest address in the account,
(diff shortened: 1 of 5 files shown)


src/monitor.rs
```diff
@@ -145,7 +145,7 @@ async fn process_output(
     let client_options_ = client_options.clone();
 
     if address_wrapper == latest_address {
-        account_handle.generate_address_internal(&mut account).await?;
+        let _ = account_handle.generate_address_internal(&mut account).await;
```
melatron, line 150: Why is the error handling removed? Maybe a comment would be nice.

src/account/sync/mod.rs
```diff
@@ -1589,7 +1589,7 @@ async fn perform_transfer(
 
     for address in addresses_to_watch {
         // ignore errors because we fallback to the polling system
-        let _ = crate::monitor::monitor_address_balance(account_handle.clone(), &address);
+        let _ = crate::monitor::monitor_address_balance(account_handle.clone(), &address).await;
```
melatron, line 1592: Why is `let _ =` needed? It is not returning a Future any more.

src/account/mod.rs
```diff
@@ -522,11 +522,19 @@ impl AccountHandle {
...
+        // monitor on a non-async function to prevent cycle computing the `monitor_address_balance` fn type
+        self.monitor_address(address.address().clone());
 
         Ok(address)
     }
 
+    pub(crate) fn monitor_address(&self, address: AddressWrapper) {
+        let handle = self.clone();
+        crate::spawn(async move {
+            let _ = crate::monitor::monitor_address_balance(handle, &address).await;
```
melatron, line 535: Why is `let _` = needed?

src/account/sync/mod.rs
lucasfernog: this one has a comment in there :P

src/monitor.rs
lucasfernog: you're right, done

src/account/mod.rs
lucasfernog: added a comment

melatron (approved):

PR #457, opened 2021-03-22
Refactoring

melatron:
# Description of change

 * Using replace for Options.
 * Removed unused slip10 crate.
 * Using map instead match.
 * Added DEFAULT_GAP_LIMIT const.

## Type of change

## Change checklist


src/account/sync/mod.rs
```diff
@@ -1620,24 +1614,12 @@ async fn is_dust_allowed(
...
-            // remove consumed outputs
-            false => {
-                if output.0 >= DUST_ALLOWANCE_VALUE {
-                    dust_allowance_balance -= output.0 as i64;
-                } else {
-                    dust_outputs_amount -= 1;
-                }
-            }
+    for (dust, add_outputs) in outputs {
+        let balance_to_add = if add_outputs { 1 } else { -1 };
```
Thoralf-M: Please rename `balance_to_add`, because it's not the balance
(if we update this part here we should also do it in iota.rs)

Thoralf-M (approved):

PR #471, opened 2021-03-29
refactor(storage): use RocksDB instead of SQLite

lucasfernog:
# Description of change

Change the database used by the wallet.

N/A

## Type of change

CLI wallet, Faucet, tests.

## Change checklist

- [x] I have made corresponding changes to the documentation


.github/workflows/build-and-test.yml
```diff
@@ -28,6 +28,13 @@ jobs:
       with:
         toolchain: stable
         override: true
+      
+    - name: Remove msys64 # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
+      if: runner.os == 'Windows'
+      run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
+    - name: Install dependencies
+      if: runner.os == 'Windows'
+      run: choco install llvm -y
```
rajivshah3: It might be worth using [KyleMayes/install-llvm-action](https://github.com/KyleMayes/install-llvm-action) here (like the Java binding does). For some reason it's much faster than using `choco` and removing the `msys64` folder

Cargo.toml
```diff
@@ -21,7 +21,7 @@ bee-common = { git = "https://github.com/iotaledger/bee.git", branch = "dev" }
 url = { version = "2.2", features = [ "serde" ] }
 tokio = { version = "1.3", features = ["macros", "sync", "time", "rt", "rt-multi-thread"] }
 rand = "0.8"
-rusqlite = { version = "0.24", features = ["bundled"] }
+rocksdb = "0.15"
```
rajivshah3: By default, it will build for all supported [compression algorithms](https://github.com/rust-rocksdb/rust-rocksdb#compression-support). Should we choose one and disable the rest?

.github/workflows/build-and-test.yml
rajivshah3: Addressed in #489

Cargo.toml
lucasfernog: do you know which one we should use? i think zstd is pretty good, but is it the best one?

rajivshah3: It seems like they [recommend](https://github.com/facebook/rocksdb/wiki/Compression#configuration) LZ4 or Zstd. Zstd gives a better compression ratio but does not compress as fast as LZ4 (probably not a big deal)

src/account/sync/mod.rs
```diff
@@ -880,10 +880,12 @@ impl AccountSynchronizer {
                 let new_addresses = data.addresses;
 
                 if !self.skip_persistence {
-                    account.append_addresses(new_addresses.to_vec());
-                    account.append_messages(parsed_messages.to_vec());
-                    account.set_last_synced_at(Some(chrono::Local::now()));
-                    account.save().await?;
+                    if !(new_addresses.is_empty() && parsed_messages.is_empty()) {
```
Thoralf-M: Maybe use `if !new_addresses.is_empty() || !parsed_messages.is_empty()` instead?

src/account_manager.rs
```diff
@@ -739,32 +732,28 @@ impl AccountManager {
...
-                return Err(crate::Error::StorageDoesntExist);
+                destination
             };
-            Ok(destination)
+            let res = fs::copy(storage_path, &destination);
+
+            let mut stronghold_storage = crate::storage::stronghold::StrongholdStorageAdapter::new(
+                &self.storage_folder.join(STRONGHOLD_FILENAME),
+            )
+            .unwrap();
```
Thoralf-M, line 744: Can't this fail?

lucasfernog: the stronghold adapter never fails; i'll add a note in there

Cargo.toml
lucasfernog: i'll go with lz4 since speed is more important here IMO

Thoralf-M (approved):

Issue #481, opened 2021-03-30
Use milestone to determine message timestamps when an account is synced from an offline state


cvarley100:
## Description

When pulling in historical messages, the milestone which confirmed a particular message should be used to determine its timestamp. From Firefly's perspective this should occur when importing an existing seed or when logging in (i.e. any time historical messages are pulled in).

## Motivation

Currently, when pulling in historical messages (on login or on first sync), the library will assign the current timestamp to the message. This is incorrect and confusing to users. The timestamp should reflect the correct time the transaction was sent/confirmed.

## Are you planning to do it yourself in a pull request?

No.

PR #495, opened 2021-04-05
Use milestone to determine message timestamps

melatron:
# Description of change

Use milestone to determine message timestamps.

Fixes issue #481

## Type of change

## Change checklist


src/message.rs
```diff
@@ -942,13 +942,33 @@ impl<'a> MessageBuilder<'a> {
             None => None,
         };
 
+        let mut timestamp = Utc::now();
+        let client_guard = crate::client::get_client(self.client_options).await?;
+        let client = client_guard.read().await;
+        let metadata = client.get_message().metadata(&self.id).await;
+        if metadata.is_ok() {
```
Thoralf-M: Why don't use `if let Ok(metadata) = client.get_message().metadata(&self.id).await{}`?

src/message.rs
```diff
@@ -942,13 +942,33 @@ impl<'a> MessageBuilder<'a> {
...
+            timestamp = match metadata.unwrap().referenced_by_milestone_index {
+                Some(ms_index) => {
+                    let ms = client.get_milestone(ms_index).await;
```
lucasfernog: @melatron can you add a cache mechanism in here? ideally we don't want to query the milestone API for each milestone index

assume you queried a milestone with index 464
and the next message was referenced by milestone 476
we can assume that the second message's timestamp is `first_message_timestamp + (476-464)*10s` since timestamps are issued every 10 seconds

of course that's not precise, so we can use that mechanism just on an specific interval, e.g. 100 indexes (so if another message has index 598, we query the node again and cache that info)

melatron: This sounds like a good optimisation I will implement it.

lucasfernog (approved):

PR #541, opened 2021-04-12
refactor(account): separate address space for change addresses

lucasfernog:
Account address space is now clearly split in public and change addresses: we'll sync them separately and always keep the latest change address unused, and use it on the transfer instead of the old logic that involved matching index with the remainder, and using latest public.

### DIFF
--- a/src/account/sync/mod.rs
+++ b/src/account/sync/mod.rs
@@ -264,10 +264,7 @@ async fn sync_address_list(
         .expect("failed to sync addresses");
     for res in results {
         let (messages, address) = res?;
-        // if the address is a change address and has no outputs, we ignore it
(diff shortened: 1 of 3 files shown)


src/account/sync/mod.rs
```diff
@@ -544,7 +538,14 @@ async fn perform_sync(
                 )
                 .await?
             } else {
-                sync_addresses(&account, address_index, gap_limit, options).await?
+                let (found_public_addresses, mut messages) =
```
Thoralf-M, line 541: Why don't you also make found_public_addresses mutable here?

lucasfernog: just for clarity on the variable name

cvarley100 (approved):

PR #562, opened 2021-04-16
feat(manager): add account discovery threshold

lucasfernog:
# Description of change

Implements `account_discovery_threshold` (`accountDiscoveryThreshold` field for actor consumers), an option on `AccountManager#sync_accounts` that determines the minimum number of accounts to check on account discovery (defaults to 1).

## Type of change

CLI wallet.

### DIFF
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -235,7 +235,7 @@ impl AccountInitialiser {
             if latest_account.with_messages(|messages| messages.is_empty()).await
                 && latest_account.addresses().iter().all(|a| a.outputs.is_empty())
             {
-                return Err(crate::Error::LatestAccountIsEmpty);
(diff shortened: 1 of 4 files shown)


src/account/mod.rs
```diff
@@ -235,7 +235,7 @@ impl AccountInitialiser {
             if latest_account.with_messages(|messages| messages.is_empty()).await
                 && latest_account.addresses().iter().all(|a| a.outputs.is_empty())
             {
-                return Err(crate::Error::LatestAccountIsEmpty);
+                // return Err(crate::Error::LatestAccountIsEmpty);
```
Thoralf-M: Why commented?

lucasfernog: oops

Thoralf-M (approved):

PR #583, opened 2021-04-22
bindings/java: Fixed milestone type, added pow fields

kwek20:
# Description of change

Updated bindings to have the new milestone changes

Not sure if the change in `src/message.rs ` is really needed, but couldnt find another way... derp


kwek20:
@lucasfernog

lucasfernog (approved):

PR #597, opened 2021-04-28
(bindings/java) Updated MigrationEvent

kwek20:
Was broken, no longer is broken ⚡

PR #613, opened 2021-05-07
use selected inputs

Thoralf-M:
# Description of change

The selected inputs weren't used, that works fine if it would only create a single bundle, but if getMigrationBundles() returns multiple input_batches we would use all inputs multiple times

## Type of change

Couldn't test with multiple inputs

## Change checklist

### DIFF
--- a/bindings/nodejs/examples/8-migration.js
+++ b/bindings/nodejs/examples/8-migration.js
@@ -75,7 +75,7 @@ async function run() {
           logFileName: 'iota-migration.log',
           // if the input is a spent address we do a bundle mining process which takes 10 minutes to reduce the amount 
           // of the parts of the private key which get revealed
-          mine: inputs.inputs[0].spent
+          mine: batch.inputs[0].spent
         })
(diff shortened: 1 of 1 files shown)


bindings/nodejs/examples/8-migration.js
```diff
@@ -71,7 +71,7 @@ async function run() {
     // create bundles with the inputs
     for (inputs of input_batches) {
       try {
-        const bundle = await manager.createMigrationBundle(seed, migrationData.inputs.map(input => input.index), {
+        const bundle = await manager.createMigrationBundle(seed, inputs.map(input => input.index), {
```
laumair: It should be `inputs.inputs` as `inputs` is an object which a an `inputs` array embedded inside. Perhaps, it is clearer if we rename `inputs` to `batch` so it will become `batch for input_batches` and then we can access it `batch.inputs`

laumair (approved):

kwek20:
@Thoralf-M I updated my other PR, should be good if you merge that first

https://github.com/iotaledger/wallet.rs/pull/597

PR #638, opened 2021-06-01
improve syncing speed and add logger example

### LINKED ISSUE #635: Improvements in syncing process
## Description

Make improvements in the syncing process to avoid unnecessary delays with a large address set. 

## Motivation

Syncing with a large address history is extremely slow. Part of it is related to the number of round trips the library makes for pulling in new history from the network. 

## Requirements

1. **Decouple address generation from syncing process:**  If you try generating new address, the polling will block this action until it completes. Address generation should be decoupled from polling;
2. **Allow users to disable background polling:** The users should be able to turn off polling if they want to manually sync in their implementations. One caveat for users (if they disable), is that they will then need to sync explicitly to detect balance changes;
3. **Only fetch outputs for addresses with positive / new balance:** Currently, we do not use `balance` endpoint and fetch outputs directly for all addresses. Instead, we should use the `balance` endpoint and fetch outputs only for the addresses that have new / positive balance.

Thoralf-M:
# Description of change

Get address balance first and only call the outputs endpoint when the address balance is not 0 or we have outputs marked as unspent
Don't return addresses without outputs in the synced account

Fixes https://github.com/iotaledger/wallet.rs/issues/635

## Type of change

## Change checklist

### DIFF
--- a/examples/logger.rs
+++ b/examples/logger.rs
@@ -1,0 +1,42 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example logger --release
+
+use iota_client::common::logger::{logger_init, LoggerConfig, LoggerOutputConfigBuilder};
+use iota_wallet::account_manager::AccountManager;
+use log::LevelFilter;
+use std::time::Instant;
+
+#[tokio::main]
+async fn main() -> iota_wallet::Result<()> {
+    // Generates a wallet.log file with logs for debugging
+    let output_config = LoggerOutputConfigBuilder::new()
+        .name("wallet.log")
+        .level_filter(LevelFilter::Debug);
+    let config = LoggerConfig::build().with_output(output_config).finish();
+    logger_init(config).unwrap();
+
+    let mut manager = AccountManager::builder()
+        .with_storage("./backup", None)?
+        .with_skip_polling()
+        .finish()
+        .await?;
+    manager.set_stronghold_password("password").await?;
+
+    let account = manager.get_account("Alice").await?;
+
+    let now = Instant::now();
+    account.sync().await.execute().await?;
+    println!("Syncing took: {:.2?}", now.elapsed());
+
+    println!("Balance: {:?}", account.balance().await?);
+
+    let addresses = account.list_unspent_addresses().await?;
+    println!("Addresses: {}", addresses.len());
+
+    let address = account.generate_address().await?;
+    println!("Generated a new address: {:?}", address);
+
+    Ok(())
+}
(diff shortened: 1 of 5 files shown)


src/account_manager.rs
```diff
@@ -1522,6 +1509,7 @@ async fn poll(
     account_options: AccountOptions,
     automatic_output_consolidation: bool,
 ) -> crate::Result<PollResponse> {
+    let polling_start_time = std::time::Instant::now();
```
melatron, line 1512: Is there a way to declare the variables needed for logging, only when we need them? Maybe hide them behind a logging feature? Not sure what is the correct way.

src/account/sync/mod.rs
```diff
@@ -1042,28 +1065,17 @@ impl AccountSynchronizer {
...
-                                .iter()
-                                .find(|(addr, _, _)| addr == &a.address().to_bech32())
-                            {
-                                Some((_, balance, outputs)) => *balance != a.balance() || outputs != a.outputs(),
-                                None => true,
-                            }
-                        })
-                        .cloned()
-                        .collect(),
+                    addresses: new_addresses,
```
melatron, line 1075: Is there any disadvantages in removing the filter? We would sync more addresses, or?

src/account/sync/mod.rs
```diff
@@ -963,6 +985,7 @@ impl AccountSynchronizer {
     /// associated with an account is fetched from the tangle and is stored locally.
     pub async fn execute(self) -> crate::Result<SyncedAccount> {
         self.account_handle.disable_mqtt();
+        let syc_start_time = std::time::Instant::now();
```
melatron, line 988: Same as above.

examples/logger.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut manager = AccountManager::builder()
+        .with_storage("./backup", None)?
+        .with_skip_polling()
+        .finish()
+        .await?;
+    manager.set_stronghold_password("password").await?;
+
+    let account = manager.get_account("Alice").await?;
+
+    let now = Instant::now();
```
melatron, line 29: Is there a way to declare the variables needed for logging, only when we need them? Maybe hide them behind a logging feature? Not sure what is the correct way.

melatron (approved):

src/account_manager.rs
Thoralf-M: Then we would need to put a feature also to the logs and for the bindings they can't set features, so I don't think that's possible in a useful way

src/account/sync/mod.rs
Thoralf-M: At this point we only return the addresses, syncing happened already before and the filtered addresses should return almost the same addresses as `new_addresses`, I think now we just get the addresses without changes but with balance added

examples/logger.rs
Thoralf-M: This example is only for logging, but apart from that I don't see a way how to do it in a useful way as I wrote in the other comment

PR #683, opened 2021-07-07
Ledger/display address and add event with PreparedTransactionData

Thoralf-M:
# Description of change

Display new receive and remainder address with ledger accounts so the user can verify them
Also added PreparedTransactionData event
Merge get_ledger_status and get_ledger_opened_app

https://github.com/iotaledger/firefly/issues/1238

## Type of change

Generated new addresses and saw the prompt on the ledger

## Change checklist


src/signing/ledger.rs
```diff
@@ -60,27 +60,24 @@ impl super::Signer for LedgerNanoSigner {
...
+        let app = match iota_ledger::get_opened_app(&transport_type) {
+            Ok((name, version)) => Some(crate::LedgerApp { name, version }),
+            _ => None,
+        };
+
+        log::info!("get_ledger");
+        let (connected_, locked) =
+            match iota_ledger::get_ledger(crate::signing::ledger::HARDENED, is_simulator).map_err(Into::into) {
+                Ok(_) => (true, false),
+                Err(crate::Error::LedgerDongleLocked) => (true, true),
```
shufps, line 73: I think `get_ledger` never returns `LedgerDongleLocked`. There are commands that are allowed although the Ledger is connected. But there are commands that could be executed afterwards to figure out if it's locked. For instance the `get_data_buffer_state` call that always can be executed is only allowed if the ledger is not locked. I think, that would be a solid test.

shufps: This call should return if the ledger is locked or not:

https://github.com/iotaledger/ledger.rs/commit/4d0656cedb744686fdb342ede15ce47ec92c532d#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R271

Thoralf-M:
@kwek20 could you please look into this `TransferProgressType` enum in java_glue.rs?
```
    /// Generating remainder value deposit address.
    GeneratingRemainderDepositAddress(AddressData),
    /// Prepared transaction.
    PreparedTransaction(PreparedTransactionData),
```
This fields hold a struct now and I didn't see how one can map it there

src/account/mod.rs
```diff
@@ -665,7 +667,39 @@ impl AccountHandle {
...
+                #[cfg(feature = "ledger-nano-simulator")]
+                SignerType::LedgerNanoSimulator => true,
+                _ => false,
+            };
+            if ledger {
+                // Send address event so it can be displayed before and then compared with the prompt on the ledger
+                emit_ledger_address_generation(&account, address.address().to_bech32()).await;
+
+                log::debug!("get_unused_address regenerate address so it's displayed on the ledger");
+                let regenrated_address = crate::address::get_address_with_index(
```
thibault-martinez: Typo regenrated_address

src/signing/stronghold.rs
```diff
@@ -28,12 +28,12 @@ pub(crate) async fn stronghold_path(storage_path: &Path) -> crate::Result<PathBu
...
     async fn get_ledger_status(&self, _is_simulator: bool) -> crate::LedgerStatus {
-        // dummy ledger status
-        crate::LedgerStatus::Connected
-    }
-
-    async fn get_ledger_opened_app(&self, _is_simulator: bool) -> crate::Result<crate::LedgerAppInfo> {
-        Err(crate::Error::NoLedgerSignerError)
+        // dummy status
+        crate::LedgerStatus {
+            connected: false,
```
melatron, line 33: shouldn't the default dummy status be connected: true?

src/event.rs
```diff
@@ -702,6 +768,40 @@ pub(crate) async fn emit_address_consolidation_needed(account: &Account, address
     }
 }
 
+/// Listen to `ledger address generation` events.
+#[cfg(any(feature = "ledger-nano", feature = "ledger-nano-simulator"))]
+#[cfg_attr(docsrs, doc(cfg(any(feature = "ledger-nano", feature = "ledger-nano-simulator"))))]
+pub async fn on_ledger_address_generation<F: Fn(&LedgerAddressGeneration) + Send + 'static>(cb: F) -> EventId {
+    let mut l = ledger_address_generation_listeners().lock().await;
```
melatron: It would be more consistent to name `l` as `listeners`.

src/signing/stronghold.rs
Thoralf-M: I would rather say no, because it's not a ledger, but at the end it shouldn't be used at all, it's only required for the trait

src/event.rs
```diff
@@ -90,6 +90,18 @@ pub struct AddressConsolidationNeeded {
...
 
+#[cfg(any(feature = "ledger-nano", feature = "ledger-nano-simulator"))]
+/// Ledger generate address event data.
+#[derive(Clone, Debug, Getters, Serialize, Deserialize)]
+#[getset(get = "pub")]
+pub struct LedgerAddressGeneration {
+    #[serde(rename = "accountId")]
+    /// The associated account identifier.
+    pub account_id: String,
+    /// The transfer event type.
```
melatron, line 101: is `addressData` the event type? The naming is kind of misleading

melatron (approved):

src/event.rs
Thoralf-M: The Event type is LedgerAddressGeneration or TransferProgressType, I will update to comment for the struct

PR #696, opened 2021-07-20
Add start and stop background sync messages to the actor

Thoralf-M:
# Description of change

Add start and stop background sync messages to the actor so it can be called from Firefly

## Type of change

## Change checklist


src/actor/message.rs
```diff
@@ -450,6 +461,13 @@ impl Serialize for MessageType {
             MessageType::GetLegacyAddressChecksum(_) => {
                 serializer.serialize_unit_variant("MessageType", 33, "GetLegacyAddressChecksum")
             }
+            MessageType::StartBackgroundSync {
+                polling_interval: _,
+                automatic_output_consolidation: _,
```
thibault-martinez: Maybe just use `..` if you're not using any field ?

src/account_manager.rs
```diff
@@ -387,7 +392,7 @@ impl Clone for AccountManager {
 
 impl Drop for AccountManager {
     fn drop(&mut self) {
-        self.stop_background_sync();
+        self.stop_background_sync().unwrap();
```
thibault-martinez, line 395: It doesn't really feel like the right place to do that to me. I would expect the user to have to call a shutdown mechanism themselves and have the opportunity to act on the error instead of panicking.

src/account_manager.rs
```diff
@@ -758,17 +763,27 @@ impl AccountManager {
...
     }
 
     /// Stops the background polling and MQTT monitoring.
-    pub fn stop_background_sync(&mut self) {
-        if let Some(polling_handle) = self.polling_handle.take() {
+    pub fn stop_background_sync(&self) -> crate::Result<()> {
+        if let Some(polling_handle) = self.polling_handle.lock().unwrap().take() {
             self.stop_polling_sender
+                .lock()
+                .map_err(|_| crate::Error::PoisonError)?
```
thibault-martinez, line 786: This is a bit inconsistent, you sometimes unwrap or map_err on the lock error. It's usually fine to always unwrap them.

src/account_manager.rs
Thoralf-M: And what would happen if the user doesn't call it before dropping it?

thibault-martinez: You tell me, I don't know what `stop_background_sync` does. But if the user doesn't call it then they are misusing the API.

Thoralf-M: It would panic then in the background syncing thread which we want to stop here so that doesn't happen

melatron (approved):

PR #751, opened 2021-08-31
update workflow

Thoralf-M:


Thoralf-M:
Electron bindings will still fail and also not sure if we need anything else from https://github.com/iotaledger/wallet.rs/tree/c88070e3e85fbf71c14d26897f1393a5d9520657/bindings/nodejs/scripts again

PR #755, opened 2021-09-02
fix(bindings/nodejs): Resolve issues with prebuilt binaries and postinstall scripts

rajivshah3:
# Description of change

Issues with `prebuild-install` and the binary not being in the right place caused the prebuilt binaries in 1.0.2 to not work properly. Also, calls made to `spawnSync` were not checked for a non-zero exit code, so the script would continue instead of aborting.

N/A

## Type of change

Tested locally

## Change checklist


Thoralf-M (approved):

laumair:
@rajivshah3 Perhaps, we could test this next release (after this PR) with the `preid` workflow? We can set it to `alpha` in the workflow and see if it works.

Issue #757, opened 2021-09-03
[Java bindings] Expose getMigrationAddress & write a reference migration example


laumair:
## Description

- Expose `getMigrationAddress` method through Java bindings. 
- Write a reference example for migrating funds from legacy to new network. Use the [nodejs example ](https://github.com/iotaledger/wallet.rs/blob/b2787949f99986ff4be3ad7912ea50aba1f5509f/bindings/nodejs/examples/8-migration.js)as a reference.

/cc @kwek20

kwek20:
This can be closed @laumair

PR #770, opened 2021-09-21
Bindings/java/migrate: Add migration example and bindings

kwek20:
# Description of change

Added an example identical to the iota.rs example

Some got mixed up with the android work, and couldnt neatly remove it, so the commit is still here. (But the android-app is not present)

Needs an updated iota.rs with the following iota-client rumqttc change: (0.8.0 doesnt do it unfortunately)
`rumqttc = { git = "https://github.com/bytebeamio/rumqtt", rev = "621a7b526936450e4e54a999eae0171bfd5bff6f", features = ["websocket"], optional = true}`
https://github.com/iotaledger/iota.rs/pull/686

fixes issue #757 

## Type of change

We want one more test with sync but require a new network: https://github.com/iotaledger/devops/issues/325

Works!

From: http://explorer.hetzner.ledgermigration1.net:3000/ledgermigration1/bundle/VS9IDTNHLHFKEHFPLKFBXILYZZABC9IJOIBZUJLANQSYGCXGZLKNCFNCVHKJTZBTAIXRVSSOV9IXQGJDZ
To: http://explorer.hetzner.ledgermigration1.net:3000/chrysalis-LedgerMigration1/addr/atoi1qqz482m2j6zyv8tkulpqzx7thqspm654k4f499eax6cmdxzhyhpq700vfkx


bindings/java/native/Cargo.toml
```diff
@@ -1,31 +1,39 @@
...
 authors = ["Brord van Wierst <brord@iota.org>"]
 edition = "2018"
 build = "build.rs"
 
 [dependencies]
-iota-wallet = { path = "../../../", version = "0.2.0" }
-iota-client = { git = "https://github.com/iotaledger/iota.rs", rev = "5f8fd262526aa09e2f548b3711964ea8fc18bc0b", features = ["mqtt"] }
-bee-message = "0.1"
+iota-wallet = { path = "../../../"}
+iota-migration = { git = "https://github.com/iotaledger/iota.rs", rev = "67193796e20841eaccc4cafb6fb442b064f8599f", package = "iota-core" }
```
laumair: Did we update it to match the latest commit?

bindings/java/native/Cargo.toml
```diff
@@ -7,7 +7,7 @@ build = "build.rs"
 
 [dependencies]
 iota-wallet = { path = "../../../", version = "0.2.0" }
-iota-client = { version = "1.1.0", default-features = false, features = ["async", "mqtt"] }
+#iota-client = { version = "1.1.0", default-features = false, features = ["async", "mqtt"] }
```
jyhi: Perhaps remove this if it doesn't have a use.

bindings/java/native/Cargo.toml
```diff
@@ -17,9 +17,17 @@ anyhow = "1.0.38"
...
 getset = { version = "0.1.1", default-features = false }
 
+# await block_on etc management
+tokio = { version = "1.12.0", default-features = false }
+once_cell = "1.4"
+futures = "0.3"
+
+# Needed for serialization
+serde = "1.0"
+serde_json = "1.0"
```
jyhi: Perhaps adding `default-features = false` as it's done above.

bindings/java/examples/basic-app/src/main/java/org/example/Migration.java
```diff
@@ -0,0 +1,330 @@
...
+     * 
+     * NOTE: If the total sum of provided inputs are less than MINIMUM_MIGRATION_BALANCE, then this method will just return and empty array as those funds can't be migrated.
+     * 
+     * This method gives precedence to max inputs over funds. It ensures, a maximum a bundle could have is 30 inputs and their accumulative balance >= MINIMUM_MIGRATION_BALANCE
+     * 
+     * @method selectInputsForUnspentAddresses
+     * 
+     * @params {List<InputData>} inputs
+     * 
+     * @returns {List<List<InputData>>}
```
jyhi: There are syntax highlighting issues; should they be `@param` and `@return` instead?

```suggestion
     * @param {List<InputData>} inputs
     * 
     * @return {List<List<InputData>>}
```

bindings/java/native/build.rs
```diff
@@ -99,4 +99,4 @@ fn gen_binding<P: AsRef<Path>>(include_dirs: &[P], c_file_path: &Path, output_ru
         .map_err(|err| err.to_string())?;
 
     Ok(())
-}
+}
```
jyhi: Not very good, there should be a newline at the end of each file.

bindings/java/native/src/classes/address.rs
```diff
@@ -40,3 +50,67 @@ impl Address {
...
+
+    /*pub fn outputs(&self, outputs: Vec<AddressOutput>) -> Self {
+        let new_builder = self
+            .builder
+            .borrow_mut()
+            .take()
+            .unwrap()
+            .outputs(outputs);
+        AddressBuilder::new_with_builder(new_builder)
+    }*/
```
jyhi: Perhaps remove it if it's useless.

bindings/java/native/src/classes/account/acc.rs
```diff
@@ -242,4 +240,49 @@ impl Account {
...
+
+    pub fn to_string(&self) -> String {
+        format!("{:?}", self.handle)
+    }
+}
+
+impl core::fmt::Display for Account {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            f, "{:?}", self.handle
```
jyhi, line 252: I don't think this is good -- just use `Debug` if someone wants to format a debug string.

bindings/java/native/src/classes/account_manager/acc_manager.rs
```diff
@@ -308,4 +300,79 @@ impl AccountManager {
             Ok(_) => Ok(()),
         }
     }
+
+    pub fn get_migration_data(&self, nodes: Vec<String>,
+        seed: &str,
+        permanode: Option<&str>,
+        security_level: i8,
+        initial_address_index: i64,
```
Thoralf-M, line 308: Can't you just use a u64 here so you don't need to add this `if initial_address_index > -1 {` check?

bindings/java/native/src/classes/address.rs
```diff
@@ -40,3 +50,67 @@ impl Address {
         self.address.address().clone()
     }
 }
+
+pub struct AddressBuilder {
+    builder: Rc<RefCell<Option<AddressBuilderRust>>>,
```
jyhi, line 66: I don't think it's wise to construct a builder pattern like this, as `RefCell` basically defers borrow checking to runtime.

See https://rust-unofficial.github.io/patterns/patterns/creational/builder.html for a reasonable structure of a builder.

jyhi (changes requested):
I don't have the condition to build and test the changes, but I do have a few comments on the code.

bindings/java/native/src/classes/account/acc.rs
kwek20: I use the Display methods to call from the toString() through java

bindings/java/native/src/classes/address.rs
kwek20: I need to add it, good spot

bindings/java/native/src/classes/address.rs
kwek20: This is the only way we can have a builder pattern through java as far as i could find. If you have suggestions please let me know :) But all calls must be through references (&self, &mut self).

bindings/java/native/src/classes/account/acc.rs
jyhi: As I see at line 244 you also have a `pub fn to_string` that formats to `{:?}`, which uses `Debug`. Is the Java `toString` bound to that function or to `Display`?

lmy441900: As I see at line 244 you also have a `pub fn to_string` that formats to `{:?}`, which uses `Debug`. Is the Java `toString` bound to that function or to `Display`?

kwek20: It just needs to be linked to another function (example: https://github.com/iotaledger/wallet.rs/blob/93a3318c1d3f952348610051b821c75b1237550a/bindings/java/native/src/java_glue.rs.in#L152)

bindings/java/native/src/classes/account_manager/acc_manager.rs
kwek20: Not supported through the bindings generator. I could add it, but this is just easier :)

Thoralf-M (approved):

jyhi (approved):

jyhi:
Our previous CI workflows don't check Java bindings, so actually this PR introduces tons of errors from `rustfmt` and `clippy`; they surface in the refreshed workflows (WIP). If @kwek20 have some time you may want to run those yourself and commit the fixes :)

PR #777, opened 2021-10-08
Fix broken links

kilianhln:
# Description of change

Found some broken links in the docs.

- Documentation Fix


Thoralf-M (approved):
Thank you!

PR #780, opened 2021-10-12
Remove double headlines

kilianhln:
# Description of change

This PR removes double headlines in the libraries example docs.

Screenshot_2

- Documentation Fix


Thoralf-M (approved):
Thanks!

Issue #743, opened 2021-08-30
[Python bindings] Improve examples / tests


laumair:
## Description

- Add more examples;
- Add missing tests.

bingyanglin:
## Testing Check List

### account_manager
- [x] start_background_sync
- [x] stop_background_sync
- [x] get_migration_data
- [x] create_migration_bundle
- [x] send_migration_bundle

### account
- [x] consolidate_outputs
- [x] generate_addresses
- [x] transfer_with_outputs

### event
- [x] on_balance_change
- [x] remove_balance_change_listener
- [x] on_new_transaction
- [x] remove_new_transaction_listener
- [x] on_confirmation_state_change
- [x] remove_confirmation_state_change_listener
- [x] on_reattachment
- [x] remove_reattachment_listener
- [x] on_broadcast
- [x] remove_broadcast_listener
- [x] on_error
- [x] remove_error_listener
- [x] on_stronghold_status_change
- [x] remove_stronghold_status_change_listener
- [x] on_transfer_progress
- [x] remove_transfer_progress_listener
- [x] on_migration_progress
- [x] remove_migration_progress_listener

bingyanglin:
## Example Check List
- [x] account_operations
- [x] backup_and_restore
- [x] consolidation
- [x] event
- [x] logger
- [x] migration
- [x] transfer

PR #790, opened 2021-10-26
feat (Python binding): Add consolidation/event examples

bingyanglin:
Solve #743

- Added consolidation example
- Added event example

### DIFF
--- a/bindings/python/examples/consolidation.py
+++ b/bindings/python/examples/consolidation.py
@@ -1,4 +1,4 @@
-# Copyright 2020 IOTA Stiftung
+# Copyright 2021 IOTA Stiftung
 # SPDX-License-Identifier: Apache-2.0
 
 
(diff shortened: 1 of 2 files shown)


bindings/python/examples/consolidation.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2020 IOTA Stiftung
```
Thoralf-M:
```suggestion
# Copyright 2021 IOTA Stiftung
```

bindings/python/examples/event.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2020 IOTA Stiftung
```
Thoralf-M:
```suggestion
# Copyright 2021 IOTA Stiftung
```

Thoralf-M (approved):

PR #793, opened 2021-11-15
Expose Gap Limit for Customization

### LINKED ISSUE #792: [Feature Request] Increase gap limit when searching for migration data
## Description

As part of legacy to chrysalis migration, we offer an [API](https://github.com/iotaledger/wallet.rs/blob/3a6a9d4c42955e6ecf84d06fffb008bb052861d8/bindings/python/native/src/classes/account_manager.rs#L117-L138) that searches for balance / transaction on a range of addresses starting from an [initialAddressIndex](https://github.com/iotaledger/wallet.rs/blob/bd5240ce0a0c7e8c8d65c3030f83374a3eec651f/src/account_manager/migration.rs#L109) and up to a [gap_limit](https://github.com/iotaledger/wallet.rs/blob/bd5240ce0a0c7e8c8d65c3030f83374a3eec651f/src/account_manager/migration.rs#L108). 

This works well for most of the cases but in some rare cases (where the funds are on a really high index), this proves cumbersome as the user will have to check again & again a lot of times to find the addresses that actually have balance. Imagine a seed having balance on an address with index 500000. 

## Motivation

A client using wallet.rs has the exact same problem as stated above and has requested to provide them a custom version of wallet.rs that has an increased `gap_limit`.

## Requirements

1. Increase the `gap_limit` to 100,000. Test it and analyse if this is a suitable number and doesn't cause any significant delays. If it does, lower the number to such that it is performant enough;
2. Test it with an example directly in wallet.rs and also with [firefly](https://github.com/iotaledger/firefly);

**Important:**

- Rather than using the latest commit, checkout from the commit used by firefly at the time when the latest version of firefly was released. The latest firefly release can be found [here](https://github.com/iotaledger/firefly/releases/tag/desktop-1.2.0). Firefly latest release pointed to `firefly-testing` branch of wallet.rs so, would be a good idea to checkout from that branch;
- After the changes, rebuild firefly locally, against the changes and test it. This [blogpost](https://blog.iota.org/firefly-token-migration/) explains how the migration process works in firefly; 
- For testing, the [firefly nodes](https://github.com/iotaledger/firefly/blob/develop/packages/shared/lib/migration.ts#L35-L36) can be used;
- Do not create a PR and keep it in a branch.  

## Open questions (optional)

N/A

No.

jyhi:
# Description of change

#792 requests a feature for an increased gap limit. It's not wise to hard-code this value, so this PR exposes it, enabling users to customize it according to their needs.

Fixes #792.

## Type of change

Enhancement (a non-breaking change which adds functionality).

```
cargo run --example migration
```

It should compile and run.

## Change checklist

### DIFF
--- a/bindings/python/native/src/classes/account_manager.rs
+++ b/bindings/python/native/src/classes/account_manager.rs
@@ -120,6 +120,7 @@ impl AccountManager {
         seed: &str,
         permanode: Option<&str>,
         security_level: Option<u8>,
+        gap_limit: Option<u64>,
         initial_address_index: Option<u64>,
     ) -> Result<MigrationData> {
         let mut finder = MigrationDataFinder::new(&nodes, seed)?;
@@ -128,6 +129,9 @@ impl AccountManager {
         }
         if let Some(initial_address_index) = initial_address_index {
             finder = finder.with_initial_address_index(initial_address_index);
+        }
+        if let Some(gap_limit) = gap_limit {
+            finder = finder.with_gap_limit(gap_limit);
         }
         if let Some(security_level) = security_level {
             finder = finder.with_security_level(security_level);
--- a/examples/migration.rs
+++ b/examples/migration.rs
@@ -32,6 +32,7 @@ async fn main() -> iota_wallet::Result<()> {
 
     // Get account data
     let mut address_index = 0;
+    let gap_limit = 30;
     let yes = vec!['Y', 'y'];
     let mut user_input = String::new();
     let mut migration_data = None;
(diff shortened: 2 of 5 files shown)


kwek20:
I noticed some gap_limit is usize, others u64. We should pick one :)

jyhi:
> some gap_limit is usize, others u64

@kwek20 They should all be u64; which usize are you referring to?

kwek20:
For example `sync_accounts` in `src/actor/mod.rs ` @lmy441900

jyhi:
Oops, indeed. `u64` is from `account_manager/migration.rs`, so there has been this inconsistency before. We should pick one!

Thoralf-M (approved):
LGTM

PR #794, opened 2021-11-18
Remove Cargo.lock

jyhi:
# Description of change

Libraries don't have their `Cargo.lock`s checked in: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

This PR removes all `Cargo.lock`s and adds it to `.gitignore`.

None.

## Type of change

Enhancement (a non-breaking change ~~which adds functionality~~).

```
cargo test
```

It should pass.

## Change checklist


rajivshah3:
While the Cargo.lock is not checked in for libraries by convention, we include it so that builds are reproducible across CI and development environments. It also helps with caching (https://github.com/iotaledger/wallet.rs/pull/258). @Thoralf-M @laumair what are your thoughts?

Thoralf-M:
For caching it should also work without it I think, but we need to change the workflows so the lock files aren't hashed there and maybe store the target folder after every build (don't know if we have limitations there)

jyhi:
Unless we do have pre-built binary users I would still propose removing it; it's the binary (executable) developers who are responsible for keeping the build reproducible (by including their `Cargo.lock` in their repository).

For workflows I will take some time to look into it to ensure nothing breaks.

Thoralf-M:
> Unless we do have pre-built binary users
For the bindings the binaries are created from our state (for nodejs and python in the workflows), so should we keep them there? For wallet.rs itself I don't see a problem with removing it

Thoralf-M (approved):

rajivshah3:
We do have people using the prebuilt binaries, so we should at least keep the lockfiles for the bindings.

Thoralf-M:
@rajivshah3 I also thought that, but at the end they will use the released version and there they can just use a fixed version?

rajivshah3:
Hmm, I guess they could. If the prebuilt binary does not work for some reason though, there's no guarantee that they will build their own binary with the same dependencies as the prebuilt one.

jyhi:
I think I need some more investigation. Before this, please don't merge.

jyhi:
I think I closed the PR by accident, but this is what I intend now.

_I did my own research (tm)_ and it turns out to be necessary to keep `Cargo.lock` in the repository and it doesn't actually cause any trouble:

- Our users won't use `Cargo.lock` anyways (neither won't it be published to crates.io).
- We publish binary build ourselves, so `Cargo.lock` is helpful for _our_ deterministic builds.
- "It also helps with caching (#258)" -- in CI workflows; including the hash of `.lock` files in the cache keys looks like a common practice.

Closing this for now (oh it's been closed, nevermind).

PR #821, opened 2021-11-30
Syncing/missing addresses

Thoralf-M:
# Description of change

Changes the syncing process so it only adds addresses that have balance and one empty address if the latest one is unused. Before when we had funds on address index 10 and synced with start index 50, which has funds, we would have a gap in the address space from 10 to 50.

## Type of change

Synced an account with different start indexes and checked the addresses afterwards

## Change checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.48"
+version = "1.0.50"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62e1f47f7dc0422027a4e370dd4548d4d66b26782e513e98dca1e689e058a80e"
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.48"
+version = "1.0.50"
```
Thoralf-M, line 49: Not sure why it shows these as changes, in dev it's already `version = "1.0.50"`

Cargo.lock
```diff
@@ -3438,4 +3438,4 @@ dependencies = [
  "quote 1.0.10",
  "syn 1.0.82",
  "synstructure",
-]
+]
```
jyhi, line 3441: Don't know what tool do this :\

jyhi (approved):

Cargo.lock
Thoralf-M: Was like this after merging the dev branch, also not sure why

Issue #825, opened 2021-12-01
list_messages  returns more messages than requested


carrie143:
let tx = await account.listMessages(1, 1, 1) should return one message,but return all recevied messages

laumair:
Related to https://github.com/iotaledger/wallet.rs/issues/769.

PR #826, opened 2021-12-02
account,storage: partially remove caching

### LINKED ISSUE #769: list_messages returns more messages than requested
## Bug description

If messages are cached, then list_messages returns more than expected if there are some cached with the same message_type, because they just get added https://github.com/iotaledger/wallet.rs/blob/6385301b1935a45da91d47dda086725fe779f597/src/account/mod.rs#L1183

## Steps To reproduce the bug

1.  
```JS
let txs = await account.listMessages(0, 0, 1)
console.log('this is length', txs.length)
txs = await account.listMessages(0, 1, 1)
console.log('this is length', txs.length)
```
`I get both prints with length 1, however, when disabling the first 2 lines, the skip in the second call is applied and i get a length of 0 (expected behaviour)`

## Expected behaviour

I don't think we should just append all cached messages
Starting with an empty Vec here https://github.com/iotaledger/wallet.rs/blob/6385301b1935a45da91d47dda086725fe779f597/src/account/mod.rs#L1183 should fix it

## Actual behaviour

Cached messages get just added

jyhi:
# Description of change

`Account::list_messages()` returns more entries than expected if
there are messages with the same `MessageType`s cached. Remove the
caching process to mitigate the problem, but
`Account::cached_messages` is kept to not break the API.

(Should) Fix #769, #825.

## Type of change

```
cargo build
```

## Change checklist

### DIFF
--- a/.changes/list-messages.md
+++ b/.changes/list-messages.md
@@ -1,0 +1,5 @@
+---
+"nodejs-binding": patch
+---
+
+Attempt to fix the wrong number of messages returned from `list_messages()`.
--- a/src/account/mod.rs
+++ b/src/account/mod.rs
@@ -6,7 +6,7 @@ use crate::{
     address::{Address, AddressBuilder, AddressOutput, AddressWrapper},
     client::{ClientOptions, Node},
     event::TransferProgressType,
-    message::{Message, MessagePayload, MessageType, TransactionEssence, Transfer},
+    message::{Message, MessageType, Transfer},
     signing::{GenerateAddressMetadata, SignerType},
     storage::{MessageIndexation, MessageQueryFilter},
 };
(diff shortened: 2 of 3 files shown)


Thoralf-M:
Please also add a change file like described here https://github.com/iotaledger/wallet.rs/tree/dev/.changes
Something likes `listMessages.md` in the .changes folder
```
---
"nodejs-binding": patch
---

Fixed returned amount with listMessages()
```

Thoralf-M (approved):

panfeilei:
The python version still suffers from this problem

Thoralf-M:
@panfeilei with the latest version from here https://nightly.link/iotaledger/wallet.rs/workflows/python_binding_publish/dev? Do you have an example to reproduce it?

PR #833, opened 2021-12-09
Adds Swift binding files and scripts

Thoralf-M:


bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[workspace]
```
jyhi: I'm not sure why this is here.

bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[package]
+name = "iota-wallet-swift-binding"
+version = "0.1.0"
+authors = ["Thoralf Müller <thoralf.mueller@iota.org", "Pasquale Ambrosini <pasquale.ambrosini@gmail.com"]
```
jyhi: Perhaps follow what we've been doing (and both of the names have the right angle bracket missing).

```suggestion
authors = ["IOTA Stiftung"]
```

bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[workspace]
+
+[package]
+name = "iota-wallet-swift-binding"
```
jyhi: I think this should stay `iota-wallet` as then it'll become the name of the output library, and we don't publish this crate anyway.

```suggestion
name = "iota-wallet"
```

bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+version = "0.1.0"
+authors = ["Thoralf Müller <thoralf.mueller@iota.org", "Pasquale Ambrosini <pasquale.ambrosini@gmail.com"]
+edition = "2018"
```
jyhi: Just go 2021 if there isn't any problem.

```suggestion
edition = "2021"
```

bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[lib]
+name = "wallet"
```
jyhi: Unless there is necessity I don't think this is needed (if we use `iota-wallet` as the name of this (bindings) crate).

bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+crate-type = ["cdylib", "staticlib"]
+
+[dependencies]
+tokio = { version = "1.11", features = ["full"] }
```
jyhi: Not very good to go for a full feature set for a library: https://docs.rs/tokio/latest/tokio/#authoring-libraries

bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+once_cell = "1.5.0"
+iota-wallet = { git = "https://github.com/iotaledger/wallet.rs", branch = "dev", default-features = false, features = ["stronghold"] }
+serde_json = "1.0"
+riker = "0.4"
+serde = "1.0"
+bee-common = "0.4.1"
```
jyhi: Please upgrade to v0.5.0 as there is security problems.

```suggestion
bee-common = "0.5.0"
```

bindings/swift/native/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+log = "0.4"
+futures = "0.3"
```
jyhi: Perhaps follow what we've been doing: `default-features = false` then pick what is really depended.

bindings/swift/native/Makefile
```diff
@@ -0,0 +1,26 @@
+
+ifeq ($(shell uname),Darwin)
+    LDFLAGS := -Wl
+	LD_LIBRARY_PATH := target/release/libwallet.dylib
```
jyhi, line 4: Please don't mix spaces and tabs :(

bindings/swift/native/Makefile
```diff
@@ -0,0 +1,26 @@
...
+else
+    LDFLAGS := -Wl,--gc-sections -lpthread -ldl
+	LD_LIBRARY_PATH := target/release/libwallet.so
+endif
+
+all: target/wallet
+	target/wallet
```
jyhi, line 11: This line is redundant.

bindings/swift/native/make_framework.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+set -e
+
+while [ $# -gt 0 ]; do
+	if [[ $1 == *"--"* ]]; then
```
jyhi, line 5: `[[` is Bash specific, but it's okay if we assume everyone uses Bash.

bindings/swift/native/make_framework.sh
```diff
@@ -0,0 +1,71 @@
...
+
+	touch $FRAMEWORK_NAME.framework/Modules/Modules.modulemap
+	tee -a $FRAMEWORK_NAME.framework/Modules/Modules.modulemap > /dev/null <<EOT
+framework module $FRAMEWORK_NAME {
+  umbrella header "$FRAMEWORK_NAME.h"
+
+  export *
+  module * { export * }
+}
+EOT
```
jyhi, line 35:
```suggestion
	cat > $FRAMEWORK_NAME.framework/Modules/Modules.modulemap << EOF
framework module $FRAMEWORK_NAME {
  umbrella header "$FRAMEWORK_NAME.h"

  export *
  module * { export * }
}
EOF
```

bindings/swift/native/make_framework.sh
```diff
@@ -0,0 +1,71 @@
...
+   <string>${FRAMEWORK_NAME}</string>
+   <key>CFBundlePackageType</key>
+   <string>FMWK</string>
+   <key>CFBundleShortVersionString</key>
+   <string>1.0</string>
+   <key>CFBundleVersion</key>
+   <string>${FRAMEWORK_VERSION}</string>
+</dict>
+</plist>
+EOT
```
jyhi, line 61: Same as above, but can't they be individual files checked into the repository? Then we just copy them.

jyhi (changes requested):
I don't have the condition to test the code, nor can I comment on the implementation, but I do have a few comments on the changes.

bindings/swift/native/make_framework.sh
pascalbros: It can't. Otherwise I need to copy the content and replace all the variables inside of it manually.
In this way we have more control and no confusion about where the templates can be.

bindings/swift/native/make_framework.sh
pascalbros: It's not the same behaviour unfortunately

jyhi: `cat >>` instead of `cat >` then.

laumair (changes requested):
I would suggest to have a dedicated section in `README` for how to use it in an iOS project. 

Also, could we perhaps add a small iOS example? We have examples for all other bindings and would be useful to have it.

PR #837, opened 2021-12-14
fix set_storage_password for messages

Thoralf-M:
# Description of change

Fix set_storage_password for messages, because we tried to read them with the new password in list_messages()

## Type of change

Changed the password for an account with messages and restored it again

## Change checklist


nicole-obrien:
Thanks for the fix Thoralf, I have tested with the latest firefly and all now seems to work as expected. 

The broken profiles are however still broken, I wonder if we can do something to recover them?

Thoralf-M:
They would be in a state with a part encrypted with the new password and a part with the old password, I would suggest just to restore the profile with a backup

nicole-obrien (approved):
LGTM and successfully tested in firefly.

PR #839, opened 2021-12-16
Drop mutex before calling finder.finish() to allow parallel executions

Thoralf-M:
# Description of change

Drop mutex before calling finder.finish() to allow parallel executions

## Type of change

Requested migration data with multiple threads, with this change it executed a lot faster and I also checked the logs, that it really executed in parallel

## Change checklist


jyhi (approved):
_Faster faster faster!_

PR #868, opened 2022-02-02
Clear storage password and verify it

### LINKED ISSUE #867: Add method to clear storage password and verify it
## Description

Currently the storage password is kept in a global static value, which let it stay there for Firefly if one logs out, so we want to add a new `clear_storage_password`? function which clears it.
When setting the storage password it can happen that we don't return an error even if the password isn't wrong, because the accounts are empty, so we should probably add an additional key in the database which we can encrypt with the password to check if it's the correct one before returning Ok() with empty accounts
For that we need to set it when accounts are loaded, because otherwise if we just set it when `set_storage_password` is called, we would set it for many users when they try to login, but if they try to login with a wrong password the first time, it wouldn't work anymore with the correct one, so we might need to check the db values which might be encrypted to see if there are already accounts (maybe by size of the data or something like that)

Maybe something like this can be used to clear the password
```Rust
    let instances = INSTANCES.get_or_init(Default::default).read().await;
    if let Some(instance) = instances.get(storage_path) {
        let mut storage_manager = instance.lock().await;
        storage_manager.storage.encryption_key.take();
        Ok(())
```

## Open questions (optional)

How exactly can we do the check with the extra key without causing problems if users try the wrong password the first time with this feature added?

## Tasks

- [ ] add function to clear the storage encryption_key
- [ ] add a way to check if a provided password in set_storage_password() is correct and return an error if not

pelumi-ajayi:
# Description of change

- Add a function to clear storage encryption key, this will also unload already decrypted accounts from memory.
- When there are no accounts loaded into memory, verify provided password in `set_storage_password` by comparing generated key checksum value (KCV) with persisted KCV for correct password .
- Use account indexation metadata in DB to validate password if key checksum value is not available.

 

Closes #867 

## Type of change

## Change checklist

### DIFF
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -238,7 +238,7 @@ impl AccountManagerBuilder {
         };
 
         if let Some(storage) = storage {
-            crate::storage::set(&storage_file_path, self.storage_encryption_key, storage).await;
+            crate::storage::set(&storage_file_path, self.storage_encryption_key, storage).await?;
         }
 
         let sync_accounts_lock = Arc::new(Mutex::new(()));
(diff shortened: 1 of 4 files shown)


Thoralf-M (approved):
LGTM!

PR #874, opened 2022-02-17
Message interface

### LINKED ISSUE #864: Implement message translator / actor system in shimmer branch
## Description

We have an [implementation](https://github.com/iotaledger/wallet.rs/tree/dev/src/actor) of an actor system (need to rethink about the name) / message passing approach. This needs to be moved to the shimmer branch. 

## Motivation

As we work towards the `shimmer` network, we need to move this implementation to the [shimmer](https://github.com/iotaledger/wallet.rs/tree/shimmer-develop) branch. This is needed so that the bindings can be written for shimmer using this approach. 

## Requirements

- Move actor / message passing approach to [shimmer](https://github.com/iotaledger/wallet.rs/tree/shimmer-develop) branch;
- Document this explaining the motivation, usage so that it is easier for bindings to be written using this approach

## Open questions (optional)

Can the current implementation be improved somehow?

No.

pelumi-ajayi:
# Description of change

Added message passing interface module for bindings, `actor` module  is now renamed as `message_interface`. Added Objective-C wrapper for Swift binding.

Closes #864 

## Type of change

Basic test was done although not exhaustive because interface has not been finalised and can still change. 

## Change checklist


src/message_interface/account_method.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2020 IOTA Stiftung
```
Thoralf-M: Should be 2022, also in the other new files

src/message_interface/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+pub use message::Message;
+pub use message_handler::WalletMessageHandler;
+pub use message_type::{AccountToCreate, MessageType};
+pub use response::Response;
+pub use response_type::ResponseType;
+
+use crate::{account_manager::AccountManager, Result};
+use tokio::sync::mpsc::unbounded_channel;
+
+pub async fn create_message_handler(path: Option<String>) -> Result<WalletMessageHandler> {
```
Thoralf-M, line 21: I think we should also allow to set other manager options here, what do you think about adding Serialize to  `AccountManagerBuilder` and use that with a method `AccountManger::from_str()` or something similar? Would require some change to the SignerHandle, so we can also build it from a String

src/message_interface/message_handler.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{AccountToCreate, Message, MessageType, Response, ResponseType, WalletMessageHandler};
+    use crate::account_manager::AccountManager;
+    use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
+
+    /// The wallet actor builder.
```
Thoralf-M: Some parts still mention `actor`

src/message_interface/mod.rs
pelumi-ajayi: Right! Setting manager options here makes total sense, I'll play with some of the ideas here but then the SignerHandle changes would have to come from iota.rs.

Thoralf-M: Currently for node.js binding we have this, I guess we want something similar
https://github.com/iotaledger/wallet.rs/blob/53d0fe2c2ce8390620f6c1b1f311e46e6ba05a03/bindings/nodejs/src/types.rs#L157
Regarding iota.rs you're right, I can also do that part

pelumi-ajayi: Alright thanks, I've made other corrections

Thoralf-M: Should be possible with this https://github.com/iotaledger/iota.rs/pull/806

PR #877, opened 2022-02-21
Remove ClientOptions

Thoralf-M:
# Description of change

Remove ClientOptions, because with the change in iota.rs it's possible to use it without them

## Type of change


laumair:
For user facing API, I still think that `clientOptions` is a better option for naming.

Thoralf-M:
Sure, we can just rename it

Thoralf-M:
> For user facing API, I still think that `clientOptions` is a better option for naming.

Done here https://github.com/iotaledger/wallet.rs/commit/295e5e330ee87e43d52055ef628afedefcbcbe19

PR #903, opened 2022-02-28
Update Swift binding

pelumi-ajayi:
# Description of change

Update swift binding with events

## Type of change

- Breaking change (feature that would cause existing functionality to not work as expected)

Unit test in Xcode

## Change checklist


bindings/swift/xcode/IotaWallet/IotaWalletTests/await.swift
```diff
@@ -16,12 +22,59 @@ class SwiftAwait: XCTestCase {
...
+    func testEvents() async throws {
+        let wallet = try! IotaWallet.Wallet()
+        let expectation = self.expectation(description: "TestEvents")
+
+        try! wallet.listen([]) { (message: String?, error: Error?) in
+            print("Event received: \(message!)")
+            expectation.fulfill()
+        }
+        
+        // TODO: Create account and do something that would trigger an event
```
Thoralf-M: Do you think it would be good to have a test event specially for such a case behind  a feature flag, so we don't have to do interactions that create an event, but can just call `emit_test_event()`?

pelumi-ajayi: Yeah, but I wouldn't know where such a function would fit.

The function would ideally be in wallet.rs, but calling it from here in Swift would mean there has to be a binding that's only used for testing, maybe such function should only exist behind a feature that's only available in debug mode to avoid misuse in production.

What do you think?

Thoralf-M: Is it possible to only be available in debug mode? Definitely behind a feature, but can't do any harm anyways

> Yeah, but I wouldn't know where such a function would fit.

I would just add it to the bottom of the AccountManager, the `verify_integrity()` method also only makes sense during development and should probably behind the same feature then

pelumi-ajayi: > Is it possible to only be available in debug mode? Definitely behind a feature, but can't do any harm anyways

I found this https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions, but maybe it's fine putting it behind a feature.

> the verify_integrity() method also only makes sense during development and should probably behind the same feature then

Good point, I'll handle that as well.

pelumi-ajayi: Since Rust features don't translate to other languages, I eventually decided to go with making emit_test_event() available in debug builds only, I think it enforces the constraint that this should only be used while testing rather than in production.

src/account_manager/mod.rs
```diff
@@ -165,13 +167,22 @@ impl AccountManager {
...
     /// automatically fix?) Also checks for each account if there is a gap in an address list and no address is
     /// duplicated
     pub async fn verify_integrity(&self) -> crate::Result<()> {
         verify_integrity(self).await
     }
 
+    #[cfg(debug_assertions)]
+    /// Checks if there is no missing account for example indexes [0, 1, 3] should panic (for now, later return error,
+    /// automatically fix?) Also checks for each account if there is a gap in an address list and no address is
+    /// duplicated
```
Thoralf-M: Wrong comment
```suggestion
    /// Helper function to test events. Emits a provided event with account index 0.
```

pelumi-ajayi: Thanks, don't know how I missed that.

bindings/swift/xcode/IotaWallet/IotaWallet.xcodeproj/project.pbxproj
```diff
@@ -14,7 +14,8 @@
 		7904C23327BBDFB2001CB01B /* Wallet.m in Sources */ = {isa = PBXBuildFile; fileRef = 7904C23027BBDFB2001CB01B /* Wallet.m */; };
 		7904C23827BBDFD2001CB01B /* await.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7904C23627BBDFD2001CB01B /* await.swift */; };
 		7904C23927BBDFD2001CB01B /* callback.m in Sources */ = {isa = PBXBuildFile; fileRef = 7904C23727BBDFD2001CB01B /* callback.m */; };
-		7904C23C27BBE08D001CB01B /* libiota_wallet.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 7904C23B27BBE08D001CB01B /* libiota_wallet.dylib */; };
+		79BD1BBB27CE54810043965C /* libiota_wallet.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 79BD1BBA27CE54810043965C /* libiota_wallet.dylib */; };
```
Thoralf-M, line 17: Is this all auto generated and is this file needed? Or could we just gitignore it so we don't always get these changes?

Thoralf-M (approved):

bindings/swift/xcode/IotaWallet/IotaWallet.xcodeproj/project.pbxproj
Thoralf-M: Feel free to merge if it's good to keep the file

pelumi-ajayi: This one is auto-generated based on Xcode settings I think, I'll add gitignore later on.

PR #909, opened 2022-03-09
Use impl AccountManager/Handle{} for the functions

Thoralf-M:
# Description of change

Use impl AccountManager/Handle{} for the functions to not have the documentation and functions signatures duplicated


Thoralf-M:
Will merge this to prevent conflicts and the logic didn't change anyways

PR #910, opened 2022-03-09
feat (Python binding): Add python class for rust methods binding

bingyanglin:
- Add python class and modules for binding

### DIFF
--- a/bindings/python/native/Cargo.toml
+++ b/bindings/python/native/Cargo.toml
@@ -12,7 +12,7 @@ keywords = ["iota", "wallet", "transaction", "python"]
 categories = ["cryptography::cryptocurrencies"]
 
 [lib]
-name = "iota_wallet_core"
+name = "iota_wallet"
 crate-type = ["cdylib"]
 
 [dependencies]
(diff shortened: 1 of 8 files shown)


bindings/python/native/Cargo.toml
```diff
@@ -12,7 +12,7 @@ keywords = ["iota", "wallet", "transaction", "python"]
 categories = ["cryptography::cryptocurrencies"]
 
 [lib]
-name = "iota_wallet"
+name = "iota_wallet_core"
```
Thoralf-M: I think this should stay `iota_wallet`

bingyanglin: Agreed. Changed and pushed.

bindings/python/native/setup.py
```diff
@@ -21,7 +21,7 @@ def get_py_version_cfgs():
 
 
 setup(
-    name="iota_wallet",
+    name="iota_wallet_core",
```
Thoralf-M:
```suggestion
    name="iota_wallet",
```

Thoralf-M (approved):

PR #938, opened 2022-03-17
Add test for address generation with a mnemonic 

bingyanglin:
Solve iotaledger/iota-sdk#73 
- Use global test vector file
- Add python testing for generate deterministic addresses from the same mnemonic


Thoralf-M (approved):

tests/fixtures/test_vectors.json
```diff
@@ -0,0 +1,132 @@
+{
+    "general": {
+        "SIGNER_TYPE": "Stronghold",
+        "MNEMNONIC": "acoustic trophy damage hint search taste love bicycle foster cradle brown govern endless depend situate athlete pudding blame question genius transfer van random vast",
+        "MNEMNONIC_ADDRESS": "iota1qpswqe4v8z2cdtgc7sfj0hfneqh37lhmjgnth36mfndwcxkjrakcvpmm727"
```
Thoralf-M: This is the public address at index 1, maybe we should use another name to make this clear?

bingyanglin: Good suggestion. Modified.

PR #967, opened 2022-03-30
feat (Python binding): Python classes cleanup

bingyanglin:
- Refactory wallet and account class
- Rename message_type to message
- Clean codes
- Shrink the code size, and make it easier to add/remove/modify in the future.

### DIFF
--- a/bindings/python/native/iota_wallet/account.py
+++ b/bindings/python/native/iota_wallet/account.py
@@ -8,18 +8,18 @@ class Account:
 
     @send_message_routine
     def __str__(self):
-        message_type = {
+        message = {
             'cmd': 'GetAccount',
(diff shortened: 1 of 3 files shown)


bindings/python/native/iota_wallet/account.py
```diff
@@ -8,182 +8,105 @@ def __init__(self, alias_index, handle):
...
             }
         }
+        if data:
+            message['payload']['method']['data'] = data
 
-        return message_type
+        return message
+
+    def generate_addresses(self, amount, options=None):
+        """Generate new unused addresses.
```
Thoralf-M:
```suggestion
        """Generate new addresses.
```

Thoralf-M (approved):

Issue #949, opened 2022-03-23
Add burn/destroy methods for native tokens and outputs


Thoralf-M:
## Description

Add methods to burn/destroy native tokens, nfts, foundry and alias outputs.

## Motivation

If native tokens or the other outputs don't have any value for one and one is only interested in the amount, then it should be possible to burn/destroy the things to free up the storage deposit

## Requirements

- [ ] Burn native tokens
- [ ] Destroy nft
- [ ] Destroy foundry
- [ ] Destroy alias output (destroy foundry outputs first)

## Open questions (optional)

Should we for alias and nft outputs first check if their addresses can control other outputs and collect/sweep them first?

## Are you planning to do it yourself in a pull request?

No.

PR #971, opened 2022-04-04
Burn native token and outputs

pelumi-ajayi:
# Description of change

Burn native token with and without foundry
Related to https://github.com/iotaledger/wallet.rs/issues/949


src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,156 @@
...
+            // Create the new alias output with the same feature blocks, just updated state_index
+            let mut new_alias_output_builder = AliasOutputBuilder::new(existing_alias_output_data.amount, alias_id)?
+                .with_state_index(alias_output.state_index() + 1)
+                .with_foundry_counter(alias_output.foundry_counter())
+                .add_unlock_condition(UnlockCondition::StateControllerAddress(
+                    StateControllerAddressUnlockCondition::new(controller_address),
+                ))
+                .add_unlock_condition(UnlockCondition::GovernorAddress(GovernorAddressUnlockCondition::new(
+                    controller_address,
+                )));
```
Thoralf-M: I think burning is a state transition for the alias and that doesn't allow use to change the addresses (only possible with a governance transition)
https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#alias-output

So I think the unlock conditions should be just set to the same from the alias output we consume

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,156 @@
...
+                Output::Foundry(output) => existing_foundry_output = Some(output),
+                _ => unreachable!("We checked if it's an alias or foundry output before"),
+            });
+
+        let existing_alias_output_data = existing_alias_output_data
+            .ok_or_else(|| Error::BurningFailed("No alias output available".to_string()))?
+            .clone();
+
+        let existing_foundry_output = existing_foundry_output
+            .ok_or_else(|| Error::BurningFailed("No foundry output available".to_string()))?
```
Thoralf-M: Just to be more clear it shouldn't be `No` but `Required` or something similar, because we could have other alias or foundry outputs available, but they can just not be used to burn the provided native tokens

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,156 @@
...
+use primitive_types::U256;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Address, token ID and amount for `burn_native_token()`
+#[serde(rename_all = "camelCase")]
+pub struct BurnNativeTokenOptions {
+    /// Bech32 encoded address. Needs to be an account address. Default will use the
+    /// first address of the account
+    pub account_address: Option<String>,
```
Thoralf-M: Because of the reasons in the other comment this field should be removed

src/account/operations/transfer/high_level/burn_native_token.rs
pelumi-ajayi: Right!

src/account/operations/transfer/high_level/burn_native_token.rs
pelumi-ajayi: I don't quite understand your point here, can you explain this some more? Thanks

src/account/operations/transfer/high_level/burn_native_token.rs
pelumi-ajayi: Thanks!

src/account/operations/transfer/high_level/burn_native_token.rs
Thoralf-M: `No alias output available` sounds like the wallet has no Alias output, but it could have another one, that is unrelated to the token id

pelumi-ajayi: Oh I get, I'll correct the error message. Thank you

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -120,4 +121,84 @@ impl AccountHandle {
...
+        let existing_foundry_output = existing_foundry_output
+            .ok_or_else(|| Error::BurningFailed("Required foundry output not found".to_string()))?
+            .clone();
+
+        drop(account);
+
+        let custom_inputs = vec![existing_alias_output_data.output_id, existing_foundry_output.output_id];
+        let options = match options {
+            Some(mut options) => {
+                // TODO: Find out if there would be a reason to append to the user provided custom inputs?
```
Thoralf-M: No, custom inputs make no sense for burning

pelumi-ajayi: Thanks for the correction.

My understanding is that burning typically involves unbalanced transactions when certain outputs go into the input side but don't appear on the output side. In the case of destroying a foundry, my thinking is that the foundry and its controlling alias output would go into the input side of the transaction while only the alias would be on the output side.

What is the correct way to go about this please, thanks in anticipation.

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,204 @@
...
+                custom_inputs: Some(custom_inputs),
+                ..Default::default()
+            }),
+        };
+
+        let outputs = match existing_alias_output_data.output {
+            Output::Alias(alias_output) => {
+                // Create the new alias output with the same feature blocks, just updated state_index
+                let mut new_alias_output_builder =
+                    AliasOutputBuilder::new(existing_alias_output_data.amount, alias_id)?
```
Thoralf-M: This would need to be `existing_alias_output_data.amount+existing_foundry_output_data.amount`, because the amount can't be burned, only the native tokens can

src/account/operations/transfer/high_level/burn_native_token.rs
Thoralf-M: @pelumi-ajayi your understanding is correct, just wanted to clarify the comment, there is no good reason to add more custom provided inputs for burning the foundry

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,204 @@
...
+            let TokenScheme::Simple(foundry_simple_ts) = existing_foundry_output.token_scheme();
+            let outputs = vec![
+                Output::Alias(new_alias_output_builder.finish()?),
+                Output::Foundry(
+                    FoundryOutputBuilder::new(
+                        minimum_storage_deposit_foundry(&byte_cost_config)?,
+                        foundry_id.serial_number(),
+                        token_id.token_tag(),
+                        TokenScheme::Simple(SimpleTokenScheme::new(
+                            foundry_simple_ts.circulating_supply(),
```
Thoralf-M: This needs to stay the `minted_token()` amount, because we didn't mint more/less

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,204 @@
...
+                            foundry_simple_ts.melted_tokens() + burn_token_amount,
```
Thoralf-M: Before we just add it here, there should be a check that we even have this amount of native tokens available in the inputs, because if we would only have for example `10` native tokens but want to burn `20`, it will not be valid

src/error.rs
```diff
@@ -105,6 +105,9 @@ pub enum Error {
     /// Minting failed
     #[error("minting failed {0}")]
     MintingFailed(String),
+    /// Minting failed
```
Thoralf-M:
```suggestion
    /// Burning failed
```

src/account/operations/transfer/high_level/burn_native_token.rs
pelumi-ajayi: Oh, thanks man! Didn't know you were answering the question in my comment.

src/account/operations/transfer/high_level/burn_native_token.rs
pelumi-ajayi: Right, but since I'm not passing in inputs explicitly, does this mean `try_select_inputs` cannot detect insufficient amount for native tokens?

Thoralf-M: Ah yeah `try_select_inputs` should return an error then

src/account/operations/transfer/high_level/sweep_outputs.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        Ok((*output_id, alias_output))
+    }
+
+    pub(crate) async fn sweep_address_outputs(
+        &self,
+        address: Address,
+        remainder_address: AddressWrapper,
+    ) -> crate::Result<()> {
+        let query_parameters = vec![
```
Thoralf-M: QueryParameter needs to be changed depending on the used endpoint

src/account/operations/transfer/high_level/sweep_outputs.rs
```diff
@@ -0,0 +1,242 @@
...
+            QueryParameter::HasStorageDepositReturnCondition(false),
+        ];
+
+        let alias_output_ids = self.client.aliases_output_ids(query_parameters.clone()).await?;
+        let basic_output_ids = self.client.output_ids(query_parameters.clone()).await?;
+        let nft_output_ids = self.client.nfts_output_ids(query_parameters).await?;
+
+        let alias_outputs = self.client.get_outputs(alias_output_ids).await?;
+        let basic_outputs = self.client.get_outputs(basic_output_ids).await?;
+        let nft_outputs = self.client.get_outputs(nft_output_ids).await?;
```
Thoralf-M: I think this needs to happen recursively with the alias/nft ids of the new outputs and foundry outputs also need to be requested and destroyed

src/account/operations/transfer/high_level/sweep_outputs.rs
```diff
@@ -0,0 +1,242 @@
...
+                    .await?;
+            }
+        }
+
+        if !output_ids.is_empty() {
+            self.send_sweep_transaction(address, output_ids.drain(..), outputs.drain(..))
+                .await?;
+        }
+
+        Ok(())
```
Thoralf-M: Can you return the transaction ids of the sweeping transactions here?

pelumi-ajayi: Will do that

src/account/operations/transfer/high_level/sweep_outputs.rs
pelumi-ajayi: I thought so at first, but then I thought ownership should be transitive, so if A own B and B owns C, then A should own B right? Unless I'm missing something though

src/account/operations/transfer/high_level/sweep_outputs.rs
pelumi-ajayi: Done

src/account/operations/transfer/high_level/sweep_outputs.rs
Thoralf-M: At the end the same ed25519 address owns it, but if it's not in the unlock condition, then you wouldn't get it

pelumi-ajayi: Hmmm... I see what you mean

src/account/operations/transfer/high_level/burn_foundry.rs
```diff
@@ -0,0 +1,208 @@
...
+                // Create the new alias output with the same feature blocks, just updated amount and state_index
+                let alias_output = AliasOutputBuilder::new_with_amount(amount, alias_id)?
+                    .with_native_tokens(native_tokens)
+                    .with_state_index(alias_output.state_index() + 1)
+                    .with_state_metadata(alias_output.state_metadata().to_vec())
+                    .with_foundry_counter(alias_output.foundry_counter())
+                    .with_unlock_conditions(alias_output.unlock_conditions().clone())
+                    .with_feature_blocks(alias_output.feature_blocks().clone())
+                    .with_immutable_feature_blocks(alias_output.immutable_feature_blocks().clone())
+                    .finish()?;
```
Thoralf-M: This can be done easier now <https://github.com/iotaledger/bee/commit/6f5721322c02cde3e1a15311729a1b94370e1a39>
You can create the builder from the previous output and only update the state_index field and amount

src/account/operations/transfer/high_level/burn_foundry.rs
```diff
@@ -0,0 +1,208 @@
...
+                        // Create the new alias output with the same feature blocks, just updated amount and state_index
+                        let alias_output = AliasOutputBuilder::new_with_amount(amount, *alias_output.alias_id())?
+                            .with_native_tokens(native_tokens)
+                            .with_state_index(alias_output.state_index() + 1)
+                            .with_state_metadata(alias_output.state_metadata().to_vec())
+                            .with_foundry_counter(alias_output.foundry_counter())
+                            .with_unlock_conditions(alias_output.unlock_conditions().clone())
+                            .with_feature_blocks(alias_output.feature_blocks().clone())
+                            .with_immutable_feature_blocks(alias_output.immutable_feature_blocks().clone())
+                            .finish()?;
```
Thoralf-M: Same here

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,243 @@
...
+            // Create the new alias output with the same feature blocks, just updated amount and state_index
+            let alias_output = AliasOutputBuilder::new_with_amount(amount, alias_id)?
+                .with_native_tokens(alias_output.native_tokens().clone())
+                .with_state_index(alias_output.state_index() + 1)
+                .with_state_metadata(alias_output.state_metadata().to_vec())
+                .with_foundry_counter(alias_output.foundry_counter())
+                .with_unlock_conditions(alias_output.unlock_conditions().clone())
+                .with_feature_blocks(alias_output.feature_blocks().clone())
+                .with_immutable_feature_blocks(alias_output.immutable_feature_blocks().clone())
+                .finish()?;
```
Thoralf-M: Same here

src/account/operations/transfer/high_level/burn_foundry.rs
pelumi-ajayi: Nice, I'll check it out

Thoralf-M: Actually amount can currently not be updated @thibault-martinez is there a reason you didn't add the option for that?

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,243 @@
...
+                ))
+            }
+        };
+
+        let output: Output = output_dto.borrow().try_into()?;
+
+        Ok(output)
+    }
+
+    fn add_native_token(output: &Output, native_token: &NativeToken) -> crate::Result<Output> {
```
Thoralf-M: Outputs have `add_native_token()`, so this isn't required?

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,243 @@
...
+                    outputs.push(output);
+                    break;
+                }
+            }
+        }
+
+        Ok((custom_inputs, outputs))
+    }
+
+    fn output_with_native_tokens<'a>(
```
Thoralf-M: Outputs have `with_native_tokens()`, so this doesn't seem to be required?

src/account/operations/transfer/high_level/burn_foundry.rs
Thoralf-M: Same for alias_id

src/account/operations/transfer/high_level/burn_native_token.rs
pelumi-ajayi: The output builders have `with_native_tokens()` but not the outputs, I think converting to `OutputDto`, modifying and converting back to `Output` is more expensive but better resistant to API breakage than using the an output builder, what do you think?

Thoralf-M: > but better resistant to API breakage than using the an output builder
I don't see why it should be more resistant

pelumi-ajayi: There was a time I had to make fixes when some of the functions changed, if I remember correctly `AliasOutputBuilder::new_with_amount(amount, alias_id)?` used to be `AliasOutputBuilder::new(amount)?` or so, that's what I mean about being resistant to breakage. A flexible helper function that uses output builders will better solve the problem though.

src/account/operations/transfer/high_level/burn_foundry.rs
thibault-martinez: I guess I didn't add them because it's the `new` API but I can do that

thibault-martinez: https://github.com/iotaledger/bee/pull/1408

examples/11_burn_native_token_without_foundry.rs
```diff
@@ -0,0 +1,64 @@
...
+    secret::{stronghold::StrongholdSecretManager, SecretManager},
+    Result, U256,
+};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    // This example uses dotenv, which is not safe for use in production
+    dotenv().ok();
+    // Setup Stronghold secret_manager
+    let secret_manager = StrongholdSecretManager::builder()
```
Thoralf-M: This isn't necessary anymore, look at the other examples we only need to set the password after the initial creation

examples/11_burn_native_token_without_foundry.rs
```diff
@@ -0,0 +1,64 @@
...
+    .unwrap();
+    let token_id = TokenId::new(token_id);
+
+    // Burn some of the circulating supply
+    let burn_amount = U256::from(10);
+    let _ = account
+        .burn_native_token_without_foundry((token_id, burn_amount), None)
+        .await
+        .unwrap();
+    tokio::time::sleep(Duration::new(15, 0)).await;
```
Thoralf-M: Is this sleep just there to wait for confirmation? Then it should rather be `retry_until_included()` from iota.rs, so it actively checks for confirmation and retries if necessary

src/account/operations/transfer/high_level/burn_foundry.rs
```diff
@@ -0,0 +1,203 @@
...
+            .filter(|output_data| match &output_data.output {
+                Output::Alias(output) => output.alias_id().or_from_output_id(output_data.output_id) == alias_id,
+                Output::Foundry(output) => output.id() == foundry_id,
+                _ => false,
+            })
+            .for_each(|output_data| match &output_data.output {
+                Output::Alias(_) => existing_alias_output_data = Some(output_data),
+                Output::Foundry(_) => existing_foundry_output = Some(output_data),
+                _ => unreachable!("We checked if it's an alias or foundry output before"),
+            });
```
Thoralf-M: Wouldn't a `for of` loop be simpler here?

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,250 @@
...
+                        if !amount.is_zero() {
+                            let not_to_be_burnt_native_tokens = not_to_be_burnt_native_tokens.into_iter().cloned();
+
+                            let output =
+                                replace_output_native_tokens(&output_data.output, not_to_be_burnt_native_tokens)?;
+                            inputs_and_outputs.push((output_id, amount, output));
+                        }
+                    }
+                }
+                Output::Treasury(_) | Output::Foundry(_) => continue,
```
Thoralf-M: I know it's called `burn_native_token_without_foundry`, but that's just because to differentiate between melting? If so, I would say it should still be possible to burn with a foundry instead of melting
For a native token that got minted by a foundry, it's probably in most cases desired to use the melt way instead, but a foundry could also have native tokens, that got minted by other foundries

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,250 @@
...
+        SimpleTokenScheme, TokenId, TokenScheme, OUTPUT_COUNT_MAX,
+    },
+};
+use primitive_types::U256;
+
+use crate::account::{handle::AccountHandle, operations::transfer::TransferResult, TransferOptions};
+
+impl AccountHandle {
+    /// Function to burn native tokens with foundry
+    pub async fn burn_native_token(
```
Thoralf-M: This should be called `melt`

src/account/operations/transfer/high_level/sweep_outputs.rs
```diff
@@ -0,0 +1,412 @@
...
+                        ..Default::default()
+                    }),
+                )
+                .await?
+                .first()
+                .ok_or_else(|| crate::Error::BurningFailed("Couldn't generate an address".to_string()))?
+                .clone());
+
+            result
+        };
```
Thoralf-M: Maybe this should be an own `pub(crate) async fn generate_remainder_address()` method

tests/burn_outputs.rs
```diff
@@ -0,0 +1,287 @@
...
+    println!("account balance -> {}", serde_json::to_string(&balance).unwrap());
+
+    // Let's destroy the first alias we can find
+    let alias_id = balance.aliases.first().unwrap().clone();
+    // let alias_id: [u8; AliasId::LENGTH] =
+    //     hex::decode("bd7f195319e3e9e3bb40f7877584944d2296b4d39c78e22b3105f9ef6dd4905e")
+    //         .unwrap()
+    //         .try_into()
+    //         .unwrap();
+    // let alias_id = AliasId::new(alias_id);
```
Thoralf-M: Can this be removed?

examples/11_burn_native_token_without_foundry.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example 11_burn_native_token_without_foundry --release
```
Thoralf-M: If we have the clear separation from burning/melting (only possible with foundry) I think we don't need to add the `without_foundry` to the name

examples/11_burn_native_token_without_foundry.rs
```diff
@@ -0,0 +1,64 @@
...
+    // This example uses dotenv, which is not safe for use in production
+    dotenv().ok();
+    // Setup Stronghold secret_manager
+    let secret_manager = StrongholdSecretManager::builder()
+        .password(&env::var("STRONGHOLD_PASSWORD").unwrap())
+        .snapshot_path(PathBuf::from("wallet.stronghold"))
+        .build();
+
+    // Create the account manager with the secret_manager and client options
+    let client_options = iota_wallet::ClientOptions::new()
```
Thoralf-M: Same for the client options, also not needed anymore to set everytime

src/account/operations/transfer/high_level/burn_foundry.rs
```diff
@@ -0,0 +1,203 @@
...
+
+use crate::{
+    account::{
+        handle::AccountHandle, operations::transfer::TransferResult, types::OutputData, SyncOptions, TransferOptions,
+    },
+    Error,
+};
+
+impl AccountHandle {
+    /// Function to destroy foundry
```
Thoralf-M: Probably good to add to the comment, that this will move any native tokens to the alias output of the foundry

src/account/operations/transfer/high_level/burn_foundry.rs
```diff
@@ -0,0 +1,203 @@
...
+    pub async fn burn_foundry(
+        &self,
+        foundry_id: FoundryId,
+        options: Option<TransferOptions>,
+    ) -> crate::Result<TransferResult> {
+        log::debug!("[TRANSFER] burn_foundry");
+
+        let alias_id = *foundry_id.alias_address().alias_id();
+        let (existing_alias_output_data, existing_foundry_output_data) =
```
Thoralf-M: Does this work every time without this check?
````
When Next State is empty, condition 1 and 2 are ignored. It must hold true, that Current State(Minted Tokens) + Token Diff = Current State(Melted Tokens)
````
from https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#simple-token-scheme-validation-rules before we can destroy it

File: src/account/operations/transfer/high_level/burn_foundry.rs [hunk 28]
```diff
@@ -0,0 +1,203 @@
...
+use crate::{
+    account::{
+        handle::AccountHandle, operations::transfer::TransferResult, types::OutputData, SyncOptions, TransferOptions,
+    },
+    Error,
+};
+
+impl AccountHandle {
+    /// Function to destroy foundry
+    pub async fn burn_foundry(
```
Thoralf-M: In https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#additional-transaction-semantic-validation-rules-2 it's called `Destroy`, I think we should also use that for the names of methods that destroy outputs and only for burning of native tokens use `burn`

src/account/operations/transfer/high_level/burn_foundry.rs
```diff
@@ -0,0 +1,203 @@
...
+        let outputs = match existing_alias_output_data.output {
+            Output::Alias(alias_output) => {
+                // Amount in foundry can't be burned, only native tokens
+                let amount = existing_alias_output_data.amount + existing_foundry_output_data.amount;
+                let mut native_tokens = Vec::from_iter(alias_output.native_tokens().clone());
+                let burn_native_token_remainder = options.as_ref().map_or(false, |options| options.allow_burning);
+                if !burn_native_token_remainder {
+                    // Transfer native tokens from foundry to alias
+                    if let Output::Foundry(foundry_output) = existing_foundry_output_data.output {
+                        native_tokens.extend(foundry_output.native_tokens().clone())
```
Thoralf-M: I think this needs to use the merge() method https://github.com/iotaledger/bee/blob/146ce1011fd76f3d9f07390b002b29dd6762daf3/bee-block/src/output/native_token.rs#L89, because we could otherwise end up with multiple entries of the same token id

src/account/operations/transfer/high_level/burn_foundry.rs
```diff
@@ -0,0 +1,203 @@
...
+                    Output::Alias(alias_output) => {
+                        // Amount in foundry can't be burned, only native tokens
+                        let amount = existing_alias_output_data.amount + existing_foundry_output_data.amount;
+                        let mut native_tokens = Vec::from_iter(alias_output.native_tokens().clone());
+                        let burn_native_token_remainder =
+                            options.as_ref().map_or(false, |options| options.allow_burning);
+                        if !burn_native_token_remainder {
+                            // Transfer native tokens from foundry to alias
+                            if let Output::Foundry(foundry_output) = existing_foundry_output_data.output {
+                                native_tokens.extend(foundry_output.native_tokens().clone())
```
Thoralf-M: Same here, use the `merge()` method instead

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,250 @@
...
+            match output_data.output {
+                Output::Alias(_) | Output::Basic(_) | Output::Nft(_) => {
+                    if let Some(native_tokens) = output_data.output.native_tokens() {
+                        let mut amount = U256::from(0);
+                        let mut not_to_be_burnt_native_tokens = Vec::new();
+                        for native_token in native_tokens.iter() {
+                            if *native_token.token_id() == token_id {
+                                amount += *native_token.amount();
+                            } else {
+                                not_to_be_burnt_native_tokens.push(native_token);
```
Thoralf-M: Should probably use `add_native_token()` https://github.com/iotaledger/bee/blob/146ce1011fd76f3d9f07390b002b29dd6762daf3/bee-block/src/output/native_token.rs#L70

src/account/operations/transfer/high_level/burn_native_token.rs
```diff
@@ -0,0 +1,250 @@
...
+        inputs_and_outputs.sort_by(|a, b| b.1.cmp(&a.1));
+
+        // Select unspent outputs with token id that sum up to the required amount
+        let mut outputs = Vec::new();
+        let mut custom_inputs = Vec::new();
+
+        let mut native_token_amount_acc = U256::from(0);
+        for input_and_output in inputs_and_outputs.into_iter().take(OUTPUT_COUNT_MAX as usize) {
+            custom_inputs.push(*input_and_output.0);
+            native_token_amount_acc += input_and_output.1;
```
Thoralf-M: @thibault-martinez do we need to do a checked add here? Shouldn't be possible to get an overflow here, but not 100% sure

examples/11_burn_native_token_without_foundry.rs
pelumi-ajayi: Exactly!

src/account/operations/transfer/high_level/burn_nft.rs
```diff
@@ -0,0 +1,68 @@
...
+    output::{BasicOutputBuilder, NftId, Output, OutputId},
+};
+
+use crate::{
+    account::{handle::AccountHandle, operations::transfer::TransferResult, TransferOptions},
+    Error,
+};
+
+impl AccountHandle {
+    /// Function to mint nft.
```
Thoralf-M: Should be burn or destroy, here it says https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#consumed-outputs-2 burning 🤔

src/account/operations/transfer/high_level/burn_foundry.rs
Thoralf-M: Hm actually for NFTs it says burning https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#consumed-outputs-2, so not sure what we should use

src/account/operations/transfer/high_level/burn_nft.rs
```diff
@@ -0,0 +1,68 @@
...
+            .ok_or(Error::NftNotFoundInUnspentOutputs)?;
+
+        let nft_output = match &output_data.output {
+            Output::Nft(nft_output) => nft_output,
+            _ => unreachable!("We already checked that it's an nft output"),
+        };
+
+        let basic_output = Output::Basic(
+            BasicOutputBuilder::new_with_amount(nft_output.amount())?
+                .with_feature_blocks(nft_output.feature_blocks().clone())
```
Thoralf-M: When we burn/destory the output I don't think we should keep the feature blocks

src/account/operations/transfer/high_level/burn_nft.rs
```diff
@@ -0,0 +1,68 @@
...
+                .with_unlock_conditions(nft_output.unlock_conditions().clone())
```
Thoralf-M: Same here, should be only the unlock condition for the address

src/account/operations/transfer/high_level/sweep_outputs.rs
```diff
@@ -0,0 +1,412 @@
...
+            custom_inputs: Some(custom_inputs),
+            try_fetch_unfound_outputs: true,
+            ..Default::default()
+        });
+
+        let transfer_result = self.send(custom_outputs, transfer_options).await?;
+        match &transfer_result.message_id {
+            Some(message_id) => {
+                let _ = self.client.retry_until_included(message_id, None, None).await?;
+                let _ = self.sync(None).await?;
```
Thoralf-M: Maybe we should add such a method with the transaction id? Shouldn't be done in this PR, but if you think it would be better let's open an issue

src/account/operations/transfer/high_level/sweep_outputs.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        Ok(output_responses)
+    }
+
+    async fn fetch_address_basic_outputs(&self, address: &AddressWrapper) -> crate::Result<Vec<OutputResponse>> {
+        let query_parameters = vec![
+            QueryParameter::Address(address.to_bech32()),
+            QueryParameter::HasExpirationCondition(false),
+            QueryParameter::HasTimelockCondition(false),
+            QueryParameter::HasStorageDepositReturnCondition(false),
```
Thoralf-M: I think because of this the comment for the high level method should also say, that such outputs could get ignored

examples/11_burn_native_token_without_foundry.rs
pelumi-ajayi: Yes, but `client` is not exposed to the wallet user. I'm thinking either `retry_until_included()` can be done in `burn_native_token()` before returning or maybe a force `sync()` of pending transactions by the user.

Thoralf-M: We could also just reexport it I think, doing `retry_until_included()` internally would slow the response down

src/account/operations/transfer/high_level/burn_foundry.rs
Thoralf-M: Use destroy here

src/account/operations/transfer/high_level/burn_nft.rs
Thoralf-M: No, should stay

Thoralf-M: Comment should be "burn"

src/account/operations/transfer/high_level/burn_foundry.rs
pelumi-ajayi: Just found out it doesn't always work, only worked when all native tokens had been melted, tried burning some and melting some and that didn't work as well.

So does this mean that a foundry must melt all of its minted tokens before it can be destroyed?

Thoralf-M: I think so

pelumi-ajayi: Hmm... But does this not make it almost impossible to destroy foundries? Because in most practical situations native tokens will be in circulation out of the foundries that minted them.

Thoralf-M: I think that's true, if it wouldn't be like that, then there could be native tokens in circulation where one couldn't find the foundry for them to get the metadata

pelumi-ajayi: Right, so this means that foundries that have some of their tokens burnt rather than melted can't ever be destroyed.

Thoralf-M: Yes in this case we should just return an error I would say

pelumi-ajayi: Is there a way to detect this for a token that was burnt without the knowledge of the foundry?

Thoralf-M: No, but the protocol also doesn't allow to burn the foundry then, so if a user burns a native without the foundry, the foundry can never be destroyed

pelumi-ajayi: And as such the controlling alias can never be destroyed as well.

Thoralf-M: The alias could still be destroyed, but of course any control about the foundry would be lost then

pelumi-ajayi: But we don't want to allow that do we? The foundry can't be unlocked without its controlling alias, could a foundry still be useful for keeping metadata in this case?

Thoralf-M: Yes I would say better not allow it for not at least, if really wanted one could destroy it with a client lib or we might add something in the future

src/account/operations/transfer/high_level/sweep_outputs.rs
pelumi-ajayi: I think this should be done in a separate PR

tests/account_manager.rs
```diff
@@ -85,7 +85,7 @@ async fn different_seed() -> Result<()> {
         .finish()
         .await?;
 
-    // Generating a new account needs to return an error, because the seed from the secret_manager is different
+    // Generating a new account needs to return an error, because the seed from the signer is different
```
thibault-martinez: Is this correct ?

src/error.rs
thibault-martinez: Wrong comment

examples/11_melt_native_token.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    // This example uses dotenv, which is not safe for use in production
+    dotenv().ok();
+
+    // Create the account manager
+    let manager = AccountManager::builder().finish().await?;
+
+    // Get the account we generated with `01_create_wallet`
```
samuel-rufi, line 22: Just for understanding, this example assumes that the other example `01_create_wallet` was started before right?

src/account/operations/transaction/high_level/burning_melting/destroy_foundry.rs
```diff
@@ -0,0 +1,226 @@
...
+                vec![Output::Alias(alias_output)]
+            }
+            _ => unreachable!("We checked if it's an alias output before"),
+        };
+
+        self.send(outputs, options).await
+    }
+
+    /// Destroy all foundries in the given set `foundry_ids`
+    // TODO: allow destroying of multiple foundries of the same alias, currently only one foundry output per alias can
```
samuel-rufi, line 79: Is this something that needs to be done within this PR? Or can this be done at a later point in time?
Maybe then have a issue to not forget

src/account/operations/transaction/high_level/minting/mint_native_token.rs
```diff
@@ -211,23 +229,20 @@ impl AccountHandle {
...
+
                 log::debug!("[TRANSACTION] sent alias output");
                 if let Some(block_id) = transaction_result.block_id {
                     self.client.retry_until_included(&block_id, None, None).await?;
-                } else {
-                    self.sync_pending_transactions().await?;
                 }
-
                 // Try to get the transaction confirmed
                 for _ in 0..10 {
```
samuel-rufi, line 238: why 0..10? and not 0..11? :trollface:

examples/11_melt_native_token.rs
Thoralf-M: Yes, this is actually required for all examples

src/account/operations/transaction/high_level/burning_melting/destroy_foundry.rs
Thoralf-M: I think it's fine to do it at a later point, because one can call the method to destroy a foundry alone before calling this
Also it would delay it more and in many cases it will no be possible to destroy foundry outputs at all if the native token amount didn't get melted before
But yes, creating an issue is good

src/account/operations/transaction/high_level/minting/mint_native_token.rs
Thoralf-M: Don't know, actually I think a lower value would be fine since they should get confirmed rather fast and if there is really a problem, not sure if 5 times more or less will help

samuel-rufi (approved):

PR #973, opened 2022-04-05
Fix message deserialization for old messages

Thoralf-M:
# Description of change

Fix message deserialization for old messages that got serialized with `unlock_blocks`

## Type of change

Added test and transfer example

## Change checklist


src/message.rs
```diff
@@ -1433,3 +1433,17 @@ pub enum MessageType {
...
+        assert!(serde_json::from_str::<Message>(serialized_message).is_ok());
```
thibault-martinez: I would also check that they are actually equal, just in case there is some default action if the key is not found or whatever

thibault-martinez (approved):

Issue #883, opened 2022-02-22
[Task]: Get all outputs that can be controlled by an account during syncing


Thoralf-M:
## Description

Currently during syncing we only request basic outputs for the addresses of an account. This logic should be updated to also get alias/foundry/nft outputs that can be controlled directly or indirectly by the addresses of the account. (Indireclty means unlocked with the alias or nft address).

## Motivation

One should be able to see all outputs one can send/for balance overview and we also need them for transaction creation.

## Requirements

1. Add fields in the SyncOptions, so theses additional outputs can be synced optionally if enabled in the options
2. Get alias/nft outputs first, then foundry and other outputs that can be controlled be the alias or nft addresses
3. Add the outputs to the account and return them in the balance overview

## Open questions (optional)

How exactly should the balance overview look with the different outputs?
I think for alias/nft/foundry ids maybe just list the output id and then one can request the full output from the account with them, but the amount and native tokens should be added up to the total balance.

## Are you planning to do it yourself in a pull request?

Maybe.

PR #986, opened 2022-04-12
Update account balance

Thoralf-M:
# Description of change

Update account balance to make it clearer

Fixes https://github.com/iotaledger/wallet.rs/issues/883

## Type of change

With the examples

## Change checklist

### DIFF
--- a/src/account/operations/helpers/time.rs
+++ b/src/account/operations/helpers/time.rs
@@ -147,7 +147,6 @@ pub(crate) fn can_output_be_unlocked_forever_from_now_on(
     current_time: u32,
     current_milestone: u32,
 ) -> bool {
-    let mut can_be_unlocked_forever = Vec::new();
     if let Some(unlock_conditions) = output_data.output.unlock_conditions() {
         for unlock_condition in unlock_conditions.iter() {
             match unlock_condition {
@@ -167,14 +166,15 @@ pub(crate) fn can_output_be_unlocked_forever_from_now_on(
                         // to check all account addresses, because the associated address
                         // can only be the unlock address or the storage deposit address and not both (unless they're
                         // the same, which would mean transaction to oneself)
-                        can_be_unlocked_forever.push(
-                            output_data.address == *expiration.return_address()
-                                || account_addresses
-                                    .iter()
-                                    .any(|a| a.address.inner == *expiration.return_address()),
-                        );
+                        if output_data.address != *expiration.return_address()
+                            || !account_addresses
+                                .iter()
+                                .any(|a| a.address.inner == *expiration.return_address())
+                        {
+                            return false;
+                        };
                     } else {
-                        can_be_unlocked_forever.push(false);
+                        return false;
                     }
                 }
                 UnlockCondition::Timelock(timelock) => {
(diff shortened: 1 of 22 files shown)


src/message_interface/response_type.rs
```diff
@@ -28,8 +28,8 @@ pub enum ResponseType {
     ReadAccounts(Vec<Account>),
     /// ListAddresses
     Addresses(Vec<AccountAddress>),
-    /// ListAddressesWithBalance.
-    AddressesWithBalance(Vec<AddressWithBalance>),
+    /// ListAddressesWithUnspentOutputs.
+    AddressesWithBalance(Vec<AddressWithUnspentOutputs>),
```
thibault-martinez: Should we also change the enum variant name ?

src/account/types/mod.rs
```diff
@@ -29,32 +29,24 @@ use serde::{Deserialize, Deserializer, Serialize};
...
-    #[serde(rename = "lockedNfts")]
-    pub locked_nfts: Vec<NftId>,
     // Aliases
     pub aliases: Vec<AliasId>,
     // Foundries
     pub foundries: Vec<FoundryId>,
+    // Outputs with multiple unlock conditions and if they can currently be spent or not. If there is a
+    // [`TimelockUnlockCondition`] or [`ExpirationUnlockCondition`] this can change at any time
+    #[serde(rename = "potentialLockedOutputs")]
+    pub potential_locked_outputs: HashMap<OutputId, bool>,
```
thibault-martinez:
```suggestion
    pub potentially_locked_outputs: HashMap<OutputId, bool>,
```

src/account/operations/helpers/time.rs
```diff
@@ -134,3 +136,65 @@ pub(crate) fn can_output_be_unlocked_now(
...
+// point in the future
+pub(crate) fn can_output_be_unlocked_forever_from_now_on(
+    // We use the addresses with unspent outputs, because other addresses of the account without unspent outputs can't
+    // be related to this output
+    account_addresses: &[AddressWithUnspentOutputs],
+    output_data: &OutputData,
+    current_time: u32,
+    current_milestone: u32,
+) -> bool {
+    let mut can_be_unlocked_forever = Vec::new();
```
thibault-martinez: Why not just return when conditions are false ? If you prefer returning at the very end, I would use a single bool that I change state instead of a whole vector

src/account/operations/syncing/mod.rs
```diff
@@ -138,38 +138,52 @@ impl AccountHandle {
         let network_id = self.client.get_network_id().await?;
         let mut account = self.write().await;
         // update used field of the addresses
-        for address_with_balance in addresses_with_balance.iter() {
-            if address_with_balance.internal {
+        for address_with_unpsnet_outputs in addresses_with_unspent_outputs.iter() {
+            if address_with_unpsnet_outputs.internal {
```
thibault-martinez: typo

src/message_interface/response_type.rs
Thoralf-M: Sure, thanks for catching this

src/account/operations/helpers/time.rs
Thoralf-M: Good catch, I copied it from another place where I would then for the expiration check the address in the address unlock condition or the one in the expiration unlock condition, depending on if it's expired or not, so I couldn't return right away
But for this case here we can do it, because we're only interested in the expired state

thibault-martinez (approved):

PR #1003, opened 2022-04-19
Add first public address during account discovery if needed

Thoralf-M:
# Description of change

Add first public address during account discovery if needed, because otherwise it can happen that we don't have any public address in the account and then it panics later at other places in the wallet

## Type of change

@nicole-obrien tested it in Firefly with a wallet that panicked without this change

## Change checklist

### DIFF
--- a/src/account/sync/mod.rs
+++ b/src/account/sync/mod.rs
@@ -911,6 +911,7 @@ async fn perform_sync(
 
     let is_latest_public_address_empty = if latest_public_address_index > max_new_public_index {
         public_addresses
+            .clone()
             .max_by_key(|a| a.key_index())
             .map(|a| a.outputs().is_empty())
             .unwrap_or(false)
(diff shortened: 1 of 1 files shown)


nicole-obrien (approved):
Works on restoring account in Firefly

src/account/sync/mod.rs
```diff
@@ -998,6 +999,19 @@ async fn perform_sync(
...
     }
 
+    // If we discover the account and the first public address isn't added, we will do it here
+    if return_all_addresses && !addresses_to_save.iter().any(|a| *a.key_index() == 0 && !a.internal()) {
+        log::debug!("[SYNC] adding first public address because we're discovering this account");
+        addresses_to_save.push(
+            (*public_addresses
+                .collect::<Vec<&Address>>()
+                .first()
+                // Save to unwrap because we generate the first address during account creation
```
thibault-martinez:
```suggestion
                // Safe to unwrap because we generate the first address during account creation
```

src/account/sync/mod.rs
```diff
@@ -998,6 +999,19 @@ async fn perform_sync(
         };
     }
 
+    // If we discover the account and the first public address isn't added, we will do it here
+    if return_all_addresses && !addresses_to_save.iter().any(|a| *a.key_index() == 0 && !a.internal()) {
+        log::debug!("[SYNC] adding first public address because we're discovering this account");
+        addresses_to_save.push(
+            (*public_addresses
+                .collect::<Vec<&Address>>()
```
thibault-martinez: Do we really need to collect here ? Can't we just `next` the iterator ?

Thoralf-M: No, changed

thibault-martinez (approved):

PR #1012, opened 2022-04-25
Add signature validation

### LINKED ISSUE #1009: [Task]: Validate signatures before sending
### Task description

Before sending a transaction the signature should be validated, so we don't send transactions that will be conflicting for sure
Just copy from https://github.com/Thoralf-M/wallet-core/blob/310ed0caf828fea1be4b87bc6c67c7204d94d733/src/account/operations/transfer/sign_transaction.rs#L91 should do the job

### Requirements

Signatures get validated before transaction is sent/stored

### Acceptance criteria

N/A

### Creation checklist

Thoralf-M:
# Description of change

Add signature validation so the wallet doesn't send transactions with invalid signatures, which happened a few times with ledger users that used a wrong seed/additional passphrase.

Fixes #1009 

## Type of change

Transfer example with added test

## Change checklist

### DIFF
--- a/src/account/sync/mod.rs
+++ b/src/account/sync/mod.rs
@@ -2777,7 +2777,7 @@ fn verify_signature(
     if let Some(UnlockBlock::Signature(signature_unlock_block)) = unlock_blocks.get(index) {
         Ok(address.verify(essence_hash, signature_unlock_block)?)
     } else {
-        return Err(crate::Error::MissingUnlockBlock);
+        Err(crate::Error::MissingUnlockBlock)
     }
 }
 
--- a/src/account/sync/mod.rs
+++ b/src/account/sync/mod.rs
@@ -2759,7 +2759,7 @@ fn verify_unlock_blocks(
     mut inputs: Vec<(Input, BeeAddress)>,
 ) -> crate::Result<()> {
     // Sort inputs
-    inputs.sort_by(|a, b| a.0.cmp(&b.0));
+    inputs.sort_by(|a, b| a.0.pack_new().cmp(&b.0.pack_new()));
     let essence_hash = transaction_payload.essence().hash();
     let unlock_blocks = transaction_payload.unlock_blocks();
     for (index, (_input, address)) in inputs.iter().enumerate() {
--- a/src/account/sync/mod.rs
+++ b/src/account/sync/mod.rs
@@ -2774,17 +2774,11 @@ fn verify_signature(
     index: usize,
     essence_hash: &[u8; 32],
 ) -> crate::Result<()> {
-    let signature_unlock_block = match unlock_blocks.get(index) {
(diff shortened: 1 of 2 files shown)


src/account/sync/mod.rs
```diff
@@ -2748,13 +2754,52 @@ pub(crate) async fn repost_message(
     Ok(message)
 }
 
+fn verify_unlock_blocks(
+    transaction_payload: &TransactionPayload,
+    mut inputs: Vec<(Input, BeeAddress)>,
+) -> crate::Result<()> {
+    // Sort inputs
+    inputs.sort_by(|a, b| a.0.cmp(&b.0));
```
thibault-martinez: That looks weird, what's happening here ?

src/account/sync/mod.rs
```diff
@@ -2748,13 +2754,52 @@ pub(crate) async fn repost_message(
...
+fn verify_signature(
+    address: &BeeAddress,
+    unlock_blocks: &UnlockBlocks,
+    index: usize,
+    essence_hash: &[u8; 32],
+) -> crate::Result<()> {
+    let signature_unlock_block = match unlock_blocks.get(index) {
+        Some(unlock_block) => match unlock_block {
+            UnlockBlock::Signature(b) => b,
+            UnlockBlock::Reference(b) => match unlock_blocks.get(b.index().into()) {
```
thibault-martinez: I think get is already resolving the reference ?

src/account/sync/mod.rs
Thoralf-M: It's sorting the inputs by OutputId, because the need to be sorted by the protocol

thibault-martinez: But this is currently sorting by `(Input, BeeAddress)`

Thoralf-M: No, by `Input` because of `.0`

src/account/sync/mod.rs
Thoralf-M: This gets the `SignatureUnlock` which is referenced by the `ReferenceUnlock`, so we can verify the referenced signature with the address

src/account/sync/mod.rs
thibault-martinez: Right, I'm tired, my bad. But still, you should probably sort by serialisation (pack_new) and not rely on the bytes layout, this could break on different architectures (endianness etc)

src/account/sync/mod.rs
thibault-martinez: Sure but isn't that what `UnlockBlocks::get` does in the first place ?

Thoralf-M: Ah it does it internally, will update

thibault-martinez (approved):

PR #1019, opened 2022-04-27
Backup

### LINKED ISSUE #808: [Task]: Back up and recovery features
## Description

The database / mnemonic back up and recovery features should be implemented in the new refactoring branch of wallet library. 

See [backup](https://wiki.iota.org/wallet.rs/libraries/nodejs/api_reference#backupdestination-password);
See [importAccounts](https://wiki.iota.org/wallet.rs/libraries/nodejs/api_reference#importaccountssource)

## Motivation

These are needed for feature completion of wallet refactoring

## Requirements

This depends on https://github.com/iotaledger/wallet.rs/issues/805

No.

### LINKED ISSUE #955: [Task]: Store Signer in db
### Task description

Currently the signer is not stored in the db
```
    #[serde(default, skip_serializing, skip_deserializing)]
    signer: Option<SignerHandle>,
```
because this would also store the password from the Stronghold signer and the mnemonic of the mnemonic signer

### Requirements

Remove Stronghold password when saving to db
Clear mnemonic when saving to db
When reading from the db, for mnemonic signer return `None`, this would need to be set during the creation with `with_signer()` again, for stronghold one would only need to set the password again

### Acceptance criteria

AccountManager with 
- Ledger signer requires no further inputs after reloading from the db
- Strong signer requires to set the password again
- Mnemonic signer needs to be provided again

### Creation checklist

Thoralf-M:
# Description of change

Added options to backup and restore with Stronghold (also possible when other secret managers are used)
Stored secret_manager in db
Updated examples to use the secret_manager from the db and only provide the Stronghold password when necessary

Fixes #808 
Fixes #955 

## Type of change

Added test

## Change checklist


pelumi-ajayi (approved):

examples/02_generate_address.rs
```diff
@@ -5,34 +5,27 @@
...
-        .await?;
+    let manager = AccountManager::builder().finish().await?;
 
     // Get the account we generated with `01_create_wallet`
     let account = manager.get_account("Alice").await?;
 
+    // Set the stronghold password
+    manager
+        .set_stronghold_password(&env::var("STRONGHOLD_PASSWORD").unwrap())
+        .await?;
```
thibault-martinez, line 27: Why is this not part of the builder ?

src/account/builder.rs
```diff
@@ -155,8 +155,12 @@ impl AccountBuilder {
             }
         };
 
-        let first_public_address =
-            get_first_public_address(&self.secret_manager, SHIMMER_COIN_TYPE, account_index).await?;
+        let first_public_address = get_first_public_address(
+            &self.secret_manager,
+            self.coin_type.clone().unwrap_or_default() as u32,
```
thibault-martinez: Clone wouldn't be needed if CoinType was Copy

examples/02_generate_address.rs
Thoralf-M: Because the password can also be cleared and then we need a way to set it again, but we could also add it additionally to the builder

src/account/builder.rs
Thoralf-M: Fixed in https://github.com/iotaledger/wallet.rs/pull/1044

src/storage/manager.rs
```diff
@@ -74,14 +75,41 @@ impl StorageManager {
...
-            .await
+            .await?;
+
+        if let Some(secret_manager) = &account_manager_builder.secret_manager {
+            let secret_manager = secret_manager.read().await;
+            let secret_manager_dto = SecretManagerDto::from(&*secret_manager);
+            // Only store secret_managers that aren't SecretManagerDto::Mnemonic, because there the Seed can't be
+            // serialized, so we can't create the SecretManager again
+            match secret_manager_dto {
+                SecretManagerDto::Mnemonic(_) => {}
```
thibault-martinez, line 86: I thought we at least wanted to store the type ?

src/message_interface/message_handler.rs
```diff
@@ -126,15 +119,24 @@ impl WalletMessageHandler {
...
                     }
-                    Ok(ResponseType::ReadAccounts(accounts))
+                    Ok(ResponseType::Accounts(accounts))
                 })
                 .await
             }
+            #[cfg(feature = "stronghold")]
+            MessageType::RestoreBackup { source, password } => {
+                let res = convert_async_panics(|| async {
+                    self.restore_backup(source.to_path_buf(), password.to_string()).await
```
thibault-martinez: It would be better if `restore_backup` only took a view (`&str`) on the password because atm `to_string` makes a copy and this copy is not zeroized.

src/storage/manager.rs
Thoralf-M: Initially I thought so, but having the type there without the mnemonic doesn't help us in any way, so I decided to not store it, because it's the same as if there is no secret_manager

thibault-martinez: Makes sense 👍🏻

src/message_interface/message_handler.rs
Thoralf-M: It's zeroized inside of `restore_backup()`, should I still change it to `&str`?

thibault-martinez: I checked the function and it looks like we do not need ownership so I'd change to reference yes, reduces risks of forgetting about zeroizing something 👍🏻

Thoralf-M: Hm for the message_interface it would work, but if a user calls the function from Rust he would need to take care of calling `.zeroize()` himself then because it doesn't get moved then, is that really better?

thibault-martinez: Not sure to follow, if they call the function themselves, they already have a password to zeroize. Passing a reference or a clone doesn't change the fact that they need to zeroize it.

Thoralf-M: When `stronghold_password` is a `String` it would be like this and internally zeroized and the user doesn't have to call it manually
```Rust
    restore_manager
        .restore_backup(
            PathBuf::from("test-storage/backup_and_restore/backup.stronghold"),
            stronghold_password,
        )
        .await?;
```
If we change it to `&str`, then the user would need to call `stronghold_password.zeroize();` himself, no?
```Rust
    restore_manager
        .restore_backup(
            PathBuf::from("test-storage/backup_and_restore/backup.stronghold"),
            &stronghold_password,
        )
        .await?;

    stronghold_password.zeroize();
```

thibault-martinez: Ok I was assuming there would always be a clone/copy like in the snippet. Let's keep it that way then.

thibault-martinez: That being said, why are we even cloning the string in the snippet? Just pass the ownership.

Thoralf-M: It's because we match on &mut, Pelumi probably did that to also return the MessageType in the response, which we don't need I believe
Would be like here then https://github.com/Thoralf-M/wallet.rs/pull/2

thibault-martinez: Looks like it make things easier to me

src/message_interface/message_handler.rs
```diff
@@ -119,61 +106,78 @@ impl WalletMessageHandler {
...
+                    self.account_manager.set_stronghold_password(&password).await?;
+                    password.zeroize();
+                    Ok(Response::Ok(()))
+                })
+                .await
+            }
+            MessageType::StoreMnemonic(mnemonic) => {
                 convert_async_panics(|| async {
-                    self.account_manager.get_node_info().await.map(ResponseType::NodeInfo)
+                    self.account_manager.store_mnemonic(mnemonic.to_string()).await?;
```
thibault-martinez: Shouldn't this be zeroized ?

Thoralf-M: Yes it should, will update

Thoralf-M: Changed it so it gets moved inside, we need to add it to iota.rs there

thibault-martinez: > Changed it so it gets moved inside, we need to add it to iota.rs there

Maybe create an issue so we don't forget about this.

Thoralf-M: Done here https://github.com/iotaledger/iota.rs/issues/988

thibault-martinez (approved):

Issue #988, opened 2022-04-12
Add Foundry Metadata Option to Native Token Creation


lzpap:
## Description

Add a new field to `NativeTokenOptions` to be able to pass data that will be put into the Immutable Metadata Block of the foundry output created while minting native tokens in https://github.com/iotaledger/wallet.rs/blob/764fb8fffe7dfd1fb3647bcf556dffb023bbec15/src/account/operations/transfer/high_level/mint_native_token.rs#L73

## Motivation

So you can pass in TIP-30 compliant JSON metadata such as:
```
{
  "name": "FooCoin",
  "description": "FooCoin is the utility and governance token of FooLand, a revolutionary protocol in the play-to-earn crypto gaming field.",
  "symbol": "FOO",
  "decimals": 3,
  "url": "https://foocoin.io",
  "logoUrl": "https://ipfs.io/ipfs/QmR36VFfo1hH2RAwVs4zVJ5btkopGip5cW7ydY4jUQBrkR"
}
```

## Requirements

1. Be able to pass  immutable foundry metadata upon native token minting.

## Open questions (optional)
Would be great to add this feature to the `cli-wallet` as well, so users can just call a command `mint_native_tokens` and pass in a path to the metadata json file.

## Are you planning to do it yourself in a pull request?

No.

PR #1037, opened 2022-05-03
Add Foundry Metadata Option to Native Token Creation

thibault-martinez:
Addresses https://github.com/iotaledger/wallet.rs/issues/988

## Change checklist

### DIFF
--- a/src/account/operations/transfer/high_level/mint_native_token.rs
+++ b/src/account/operations/transfer/high_level/mint_native_token.rs
@@ -65,6 +65,7 @@ impl AccountHandle {
     ///     token_tag: TokenTag::new([0u8; 12]),
     ///     circulating_supply: U256::from(100),
     ///     maximum_supply: U256::from(100),
+    ///     foundry_metadata: None
     /// };
     ///
     /// let res = account_handle.mint_native_token(native_token_options, None,).await?;
--- a/src/account/operations/transfer/high_level/mint_native_token.rs
+++ b/src/account/operations/transfer/high_level/mint_native_token.rs
@@ -6,6 +6,7 @@ use iota_client::{
     bee_message::{
         address::{Address, AliasAddress},
         output::{
+            feature_block::{FeatureBlock, MetadataFeatureBlock},
             unlock_condition::{
                 AddressUnlockCondition, GovernorAddressUnlockCondition, ImmutableAliasAddressUnlockCondition,
                 StateControllerAddressUnlockCondition, UnlockCondition,
@@ -30,11 +31,10 @@ use crate::{
     Error,
 };
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
 /// Address and nft for `mint_native_token()`
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct NativeTokenOptions {
-    /// Bech32 encoded address. Needs to be an account address. Default will use the
-    /// first address of the account
+    /// Bech32 encoded address. Needs to be an account address. Default will use the first address of the account
     #[serde(rename = "accountAddress")]
     pub account_address: Option<String>,
     /// Token tag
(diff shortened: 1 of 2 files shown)


src/account/operations/transfer/high_level/mint_native_token.rs
```diff
@@ -46,6 +46,9 @@ pub struct NativeTokenOptions {
     /// Maximum supply
     #[serde(rename = "maximumSupply")]
     pub maximum_supply: U256,
+    /// Foundry metadata
+    #[serde(rename = "foundryMetadata")]
+    pub foundry_metadata: Vec<u8>,
```
thibault-martinez: Need to make that an option

src/account/operations/transfer/high_level/mint_native_token.rs
```diff
@@ -152,9 +155,9 @@ impl AccountHandle {
             )?])?;
 
             let outputs = vec![
-                Output::Alias(new_alias_output_builder.finish()?),
-                Output::Foundry(
-                    FoundryOutputBuilder::new_with_amount(
+                new_alias_output_builder.finish_output()?,
+                {
+                    let mut foundry_builder = FoundryOutputBuilder::new_with_amount(
                         minimum_storage_deposit_foundry(&byte_cost_config)?,
```
Thoralf-M, line 162: I think we need to change this, if there is `foundry_metadata`
And what I'm also not sure is if we should increase the storage deposit so the foundry can mint native tokens without further inputs (add the required storage deposit amount for one NativeTokens)

src/account/operations/transfer/high_level/mint_native_token.rs
```diff
@@ -46,6 +46,9 @@ pub struct NativeTokenOptions {
     /// Maximum supply
     #[serde(rename = "maximumSupply")]
     pub maximum_supply: U256,
+    /// Foundry metadata
+    #[serde(rename = "foundryMetadata")]
+    pub foundry_metadata: Option<Vec<u8>>,
 }
 
 impl AccountHandle {
```
Thoralf-M, line 54: Can't comment at line 67, but the example there needs to be updated with the foundry_metadata field

thibault-martinez: Right, it's ignored so I missed it. Fixed.

src/account/operations/transfer/high_level/mint_native_token.rs
thibault-martinez: Regarding the first sentence you are right. But in https://github.com/iotaledger/wallet.rs/pull/1007 I'm removing `minimum_storage_deposit_foundry` and using `new_with_minimum_storage_deposit` so it should take care of the metadata to compute the appropriate amount. I'm not sure what you mean with the second sentence though, can you elaborate?

Thoralf-M: For the minted native tokens we currently add a new basic output which will get them, but if the foundry would have a larger amount, we could also add the native tokens to the foundry output directly, that way we have one output less (and if we don't mint any native tokens in this call, we could the next time use only the alias and foundry output as input and mint the native tokens to the foundry output)

Thoralf-M: Just a possible optimization, not really required

Thoralf-M (approved):

src/account/operations/transfer/high_level/mint_native_token.rs
thibault-martinez: Well at least the first sentence is an actual mistake

Thoralf-M: Yes, but if it's fixed in https://github.com/iotaledger/wallet.rs/pull/1007 it should be fine

PR #1044, opened 2022-05-04
Add custom de/serialization for CoinType

Thoralf-M:
# Description of change

Add custom de/serialization for CoinType, so the bindings can provide the number for it

## Type of change

Added test

## Change checklist

### DIFF
--- a/src/account/types/mod.rs
+++ b/src/account/types/mod.rs
@@ -18,10 +18,11 @@ use iota_client::{
         MessageId,
     },
     bee_rest_api::types::responses::OutputResponse,
+    constants::{IOTA_COIN_TYPE, SHIMMER_COIN_TYPE},
     secret::types::{InputSigningData, OutputMetadata},
 };
 use primitive_types::U256;
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 /// The balance of an account, returned from [`crate::account::handle::AccountHandle::sync()`] and
 /// [`crate::account::handle::AccountHandle::balance()`].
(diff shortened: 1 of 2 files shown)


src/account/types/mod.rs
```diff
@@ -228,3 +230,58 @@ impl TryFrom<u32> for CoinType {
...
+// Custom serialization to return the coin type number
+impl Serialize for CoinType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match *self {
+            CoinType::IOTA => serializer.serialize_u32(self.clone() as u32),
+            CoinType::Shimmer => serializer.serialize_u32(self.clone() as u32),
+        }
```
Tuditi: Why use a match statement if the actions undertaken are the same for both options?

Tuditi (approved):
Looks good to me! I had a minor question out of interest.

src/account/types/mod.rs
Thoralf-M: Good point, you're right shouldn't be a match there
Initially I provided the matching coin type number hardcoded, so there it wasn't the same, but didn't remove the match after changing it, will do now

src/account/types/mod.rs
```diff
@@ -204,10 +205,11 @@ impl From<u32> for AccountIdentifier {
 }
 
 /// BIP 44 coin type
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
thibault-martinez: Make it copy since it's backed by an u32, will avoid some explicit clones

src/account/types/mod.rs
```diff
@@ -228,3 +230,55 @@ impl TryFrom<u32> for CoinType {
...
     }
 }
+
+// Custom serialization to return the coin type number
+impl Serialize for CoinType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_u32(self.clone() as u32)
```
thibault-martinez: Like this

thibault-martinez (approved):

PR #1069, opened 2022-05-16
Only add account index if not already present

Thoralf-M:
# Description of change

Only add account index if not already present, because it would otherwise lead to duplicated accounts

## Type of change

Examples

## Change checklist

### DIFF
--- a/src/storage/manager.rs
+++ b/src/storage/manager.rs
@@ -131,7 +131,11 @@ impl StorageManager {
     }
 
     pub async fn save_account(&mut self, account: &Account) -> crate::Result<()> {
-        self.account_indexes.push(*account.index());
+        // Only add account index if not already present
+        if !self.account_indexes.iter().any(|index| index == account.index()) {
+            self.account_indexes.push(*account.index());
+        }
+
         self.storage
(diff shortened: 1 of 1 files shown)


Tuditi (dismissed):
LGTM!

src/storage/manager.rs
```diff
@@ -131,7 +131,11 @@ impl StorageManager {
     }
 
     pub async fn save_account(&mut self, account: &Account) -> crate::Result<()> {
-        self.account_indexes.push(*account.index());
+        // Only add account index if not already present
+        if !self.account_indexes.iter().any(|index| index == account.index()) {
```
thibault-martinez: use `contains` instead ?

pelumi-ajayi (dismissed):

src/storage/manager.rs
Thoralf-M: Wen Clippy saying me that?

thibault-martinez: Yeah it definitely should :(

thibault-martinez (approved):

samuel-rufi (approved):

PR #1071, opened 2022-05-18
Add clear_stronghold_password() and is_stronghold_password_available()

Thoralf-M:
# Description of change

Add clear_stronghold_password() and is_stronghold_password_available()

## Type of change

Example, clear_stronghold_password() works, but actions are still possible afterwards, needs to be fixed in iota.rs

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ backtrace = { version = "0.3.62", default-features = false, features = ["std"] }
 futures = { version = "0.3.17", default-features = false }
 getset = { version = "0.1.1", default-features = false }
 hex = { version = "0.4.3", default-features = false }
-iota-client = { git = "https://github.com/iotaledger/iota.rs", rev = "ba1947626f00eb6f4d5eb986c438351b89d6dbaf", default-features = false, features = ["tls"] }
+iota-client = { git = "https://github.com/iotaledger/iota.rs", rev = "57e96cbd5a6bfa0f5af027420eaa46a71866d4b2", default-features = false, features = ["tls"] }
 # iota-client = { path = "../iota.rs", default-features = false, features = ["async", "tls"] }
 iota-crypto = { version = "0.10.0", default-features = false, features = ["std", "random", "sha", "pbkdf", "hmac", "bip39", "bip39-en", "chacha", "blake2b", "slip10"] }
 log = { version = "0.4.14", default-features = false }
--- a/src/account_manager/operations/stronghold.rs
+++ b/src/account_manager/operations/stronghold.rs
@@ -12,6 +12,7 @@ impl AccountManager {
             stronghold.set_password(password).await;
             let result = stronghold.read_stronghold_snapshot().await;
             if let Err(err) = result {
+                // TODO: replace with actual error matching when updated to the new Stronghold version
                 if let iota_client::Error::StrongholdProcedureError(ref err_msg) = err {
                     if !err_msg.contains("IOError") {
                         return Err(err.into());
--- a/src/account_manager/operations/stronghold.rs
+++ b/src/account_manager/operations/stronghold.rs
@@ -6,7 +6,7 @@ use iota_client::secret::SecretManager;
 use crate::account_manager::AccountManager;
 
 impl AccountManager {
-    /// Set the Stronghold password
+    /// Sets the Stronghold password
     pub async fn set_stronghold_password(&self, password: &str) -> crate::Result<()> {
         if let SecretManager::Stronghold(stronghold) = &mut *self.secret_manager.write().await {
             stronghold.set_password(password).await;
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -22,7 +22,7 @@ backtrace = { version = "0.3.62", default-features = false, features = ["std"] }
 futures = { version = "0.3.17", default-features = false }
 getset = { version = "0.1.1", default-features = false }
 hex = { version = "0.4.3", default-features = false }
-iota-client = { git = "https://github.com/iotaledger/iota.rs", rev = "ba1947626f00eb6f4d5eb986c438351b89d6dbaf", default-features = false, features = ["tls"] }
+iota-client = { git = "https://github.com/iotaledger/iota.rs", rev = "57e96cbd5a6bfa0f5af027420eaa46a71866d4b2", default-features = false, features = ["tls"] }
 # iota-client = { path = "../iota.rs", default-features = false, features = ["async", "tls"] }
```
thibault-martinez, line 26: Should we remove this ?

src/message_interface/response.rs
```diff
@@ -44,6 +44,8 @@ pub enum Response {
     SentTransfer(TransferResult),
     /// TryCollectOutputs and CollectOutputs response.
     SentTransfers(Vec<TransferResult>),
+    /// IsStrongholdPasswordAvailable
```
thibault-martinez: Comment is different than the variant ?

Cargo.toml
Thoralf-M: Yes, did it in the other PR already

src/message_interface/response.rs
Thoralf-M: The comment is the message type which returns this variant

thibault-martinez: Sure but can't we harmonise ?

Thoralf-M: Some other comments end with `response.`, should we rather use `Response for` before each of the message types?

thibault-martinez: Yeah that sounds good, but I would also make the message type and the response more similar, it's weird that the `Is` is not at the same position

Thoralf-M: One is a question and the other one the response :trollface: Which one do you prefer?

src/account_manager/operations/stronghold.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_client::secret::SecretManager;
+
+use crate::account_manager::AccountManager;
+
+impl AccountManager {
+    /// Set the Stronghold password
```
thibault-martinez:
```suggestion
    /// Sets the Stronghold password
```

src/account_manager/operations/stronghold.rs
```diff
@@ -0,0 +1,53 @@
...
+                    if !err_msg.contains("IOError") {
+                        return Err(err.into());
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Store a mnemonic into the Stronghold vault
```
thibault-martinez:
```suggestion
    /// Stores a mnemonic into the Stronghold vault
```

src/message_interface/response.rs
thibault-martinez: Nah it's fine, let's keep it that way

src/account_manager/operations/stronghold.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl AccountManager {
+    /// Sets the Stronghold password
+    pub async fn set_stronghold_password(&self, password: &str) -> crate::Result<()> {
+        if let SecretManager::Stronghold(stronghold) = &mut *self.secret_manager.write().await {
+            stronghold.set_password(password).await;
+            let result = stronghold.read_stronghold_snapshot().await;
+            if let Err(err) = result {
+                if let iota_client::Error::StrongholdProcedureError(ref err_msg) = err {
+                    if !err_msg.contains("IOError") {
```
thibault-martinez, line 17: Hmm I don't like this, why is it not possible to do this at type level ?

Thoralf-M: The error there is just a String, maybe can be changed in the stronghold part

thibault-martinez: It would be best yeah, because a change in type would result in a compilation error but would be silent with a String...

Thoralf-M: The String is already returned on Stronghold level https://github.com/iotaledger/stronghold.rs/blob/b7fb5e9453883b3ff23d935e4e581517c42c098c/client/src/actors/client.rs#L113 and not in iota.rs
I think with the new Stronghold version it's not like this anymore and we will update to it at some point

thibault-martinez: Then please at least leave a TODO so we don't forget about it

Thoralf-M: Added

thibault-martinez (approved):

Tuditi (approved):
LGMT!

PR #1086, opened 2022-05-25
Update and emit events

### LINKED ISSUE #952: Update and emit wallet events
## Description

We need to emit wallet events in all places where needed and add more/change some to the new layouts.

## Motivation

Some users want to do action based on events and for the Ledger Signer it's required that we emit the generated address and show the transaction data in Firefly, so the user can compare it.

## Requirements

Update and emit
1. PreparedTransactionEventData to the new transaction layout (include full outputs)
2. BalanceChangeEvent should be renamed to something like NewOutputEvent and include the output, for more information see https://github.com/iotaledger/wallet.rs/issues/895. It should also be clear if the output can be really owned (no timelock/expiration/storage deposit return unlock condition), maybe for such outputs we should create an own event type.
3. TransactionInclusionEvent (only emit, no need to change the layout)

## Open questions (optional)

Any other events that could be useful?

No.

Thoralf-M:
# Description of change

Update and emit events and update ledger address generation for it

Fixes #952 

## Type of change

## Change checklist

### DIFF
--- a/src/account/operations/address_generation.rs
+++ b/src/account/operations/address_generation.rs
@@ -83,16 +83,65 @@ impl AccountHandle {
 
         let address_range = highest_current_index_plus_one..highest_current_index_plus_one + amount;
 
-        #[cfg(all(feature = "events", feature = "ledger_nano"))]
-        // If we don't sync, then we want to display the prompt on the ledger with the address. But the user needs to
-        // have it visible on the computer first, so we need to generate it without the prompt first
-        if !options.metadata.syncing {
-            match *self.secret_manager.read().await {
-                SecretManager::LedgerNano(_) | SecretManager::LedgerNanoSimulator(_) => {
+        // If we use stronghold we need to read the snapshot in case it hasn't been done already
+        #[cfg(feature = "stronghold")]
+        if let SecretManager::Stronghold(stronghold_secret_manager) = &mut *self.secret_manager.write().await {
+            stronghold_secret_manager.read_stronghold_snapshot().await?;
+        }
+
+        let addresses = match *self.secret_manager.read().await {
+            #[cfg(feature = "ledger_nano")]
+            SecretManager::LedgerNano(_) | SecretManager::LedgerNanoSimulator(_) => {
+                // If we don't sync, then we want to display the prompt on the ledger with the address. But the user
+                // needs to have it visible on the computer first, so we need to generate it without the
+                // prompt first
+                if !options.metadata.syncing {
                     let mut changed_metadata = options.metadata.clone();
+                    // Change metadata so ledger will not show the prompt the first time
                     changed_metadata.syncing = true;
-                    let addresses = self
-                        .secret_manager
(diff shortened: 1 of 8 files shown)


src/events/types.rs
```diff
@@ -16,64 +19,65 @@ pub struct Event {
     pub event: WalletEvent,
 }
 
-// do we want an event for transaction confirmation or if it failed?
-// event for new detected outputs?
-
-#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub enum WalletEvent {
-    BalanceChange(BalanceChangeEvent),
```
thibault-martinez, line 24: Can you order them the same way as WalletEventType ?

src/account/operations/syncing/mod.rs
```diff
@@ -206,10 +223,25 @@ impl AccountHandle {
         }
 
         // Add new synced outputs
-        for output in new_outputs {
-            account.outputs.insert(output.output_id, output.clone());
-            if !output.is_spent {
-                account.unspent_outputs.insert(output.output_id, output);
+        for output_data in unspent_outputs {
+            // Inser output, if it's unknown emit the NewOutputEvent
```
thibault-martinez:
```suggestion
            // Insert output, if it's unknown emit the NewOutputEvent
```

src/account/operations/address_generation.rs
```diff
@@ -83,58 +83,91 @@ impl AccountHandle {
...
+                            .read()
+                            .await
+                            .generate_addresses(
+                                account.coin_type,
+                                account.index,
+                                address_index..address_index + 1,
+                                options.internal,
+                                options.metadata.clone(),
+                            )
+                            .await?;
```
thibault-martinez, line 139: What's different between the two generations ?

Thoralf-M: The first one will just generate the address and return it direclty, the second one shows the address on the screen of the ledger nano and requires physical approval to continue

thibault-martinez: Sorry if I'm being stupid but the two generation are exactly the same so why can't we just do the emit under the feature ?

Thoralf-M: The first one has `changed_metadata.clone(),`  where syncing is set to true, which makes the ledger to not display the address and return without user approval
Then the event with the address gets emitted and Firefly can display the address, so when the second generation is called , with `options.metadata.clone()` and the ledger displays the address, the user can compare it with the address displayed in Firefly

thibault-martinez: Ah damn I don't know how I missed changed_metadata != options.metadata, sorry about that 🤦🏻

thibault-martinez (approved):

Tuditi (approved):
LGTM!

Issue #1083, opened 2022-05-24
[Task]: Add method to delete the last account


Thoralf-M:
### Task description

Firefly wants to be able to delete the last account, it's only needed for the latest account, so the order can't be messed up

### Requirements

Add a method to delete the latest account

### Acceptance criteria

Only the latest account/account with highest index can be deleted
Db is also updated with this account removed

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

13413rahul:
hey @Thoralf-M 
I would like to do this please assign it to me

Thoralf-M:
@13413rahul great! Let me know if you need more information

Thoralf-M:
@13413rahul how is it looking? We would like to get this done this week, if you don't find time for it, we will do it

PR #1116, opened 2022-06-02
Add function to remove the latest account

samuel-rufi:
# Description of change

Introduces a function to remove the latest account (the account with the largest account index).

Fixes issue https://github.com/iotaledger/wallet.rs/issues/1083

## Type of change

Tested the change with the provided unit test.

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have checked that new and existing unit tests pass locally with my changes

### DIFF
--- a/src/account_manager/mod.rs
+++ b/src/account_manager/mod.rs
@@ -93,15 +93,13 @@ impl AccountManager {
         }
 
         if let Some(largest_account_index) = largest_account_index_opt {
-            let mut i = 0;
-            while i < accounts.len() {
+            for i in 0..accounts.len() {
                 if let Some(account) = accounts.get(i) {
                     if *account.read().await.index() == largest_account_index {
                         let _ = accounts.remove(i);
                         return Ok(());
                     }
                 }
-                i += 1;
             }
 
             #[cfg(feature = "storage")]
--- a/src/account_manager/mod.rs
+++ b/src/account_manager/mod.rs
@@ -77,7 +77,7 @@ impl AccountManager {
     }
 
     /// Removes the latest account (account with the largest account index).
-    pub async fn remove_latest_account(&mut self) -> crate::Result<()> {
+    pub async fn remove_latest_account(&self) -> crate::Result<()> {
         let mut accounts = self.accounts.write().await;
 
         let mut largest_account_index_opt = None;
--- a/src/account_manager/mod.rs
+++ b/src/account_manager/mod.rs
@@ -76,6 +76,45 @@ impl AccountManager {
         Ok(self.accounts.read().await.clone())
     }
 
+    /// Remove the latest account (account with the largest account index).
+    pub async fn remove_latest_account(&mut self) -> crate::Result<()> {
+        let mut accounts = self.accounts.write().await;
+
+        let mut largest_account_index_opt = None;
+        for account in accounts.iter() {
+            let account_index = *account.read().await.index();
+            if let Some(largest_account_index) = largest_account_index_opt {
+                if account_index > largest_account_index {
+                    largest_account_index_opt = Some(account_index);
+                }
+            } else {
+                largest_account_index_opt = Some(account_index)
+            }
+        }
+
+        if let Some(largest_account_index) = largest_account_index_opt {
+            let mut i = 0;
+            while i < accounts.len() {
+                if let Some(account) = accounts.get(i) {
+                    if *account.read().await.index() == largest_account_index {
+                        let _ = accounts.remove(i);
+                        break;
+                    }
+                }
+                i = i + 1;
+            }
+
+            #[cfg(feature = "storage")]
+                self.storage_manager
+                .lock()
+                .await
+                .remove_account(largest_account_index)
+                .await?;
+        }
+
+        Ok(())
+    }
+
     /// Get the [SecretManager]
     pub fn get_secret_manager(&self) -> Arc<RwLock<SecretManager>> {
         self.secret_manager.clone()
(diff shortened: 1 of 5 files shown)


src/account_manager/mod.rs
```diff
@@ -76,6 +76,44 @@ impl AccountManager {
         Ok(self.accounts.read().await.clone())
     }
 
+    /// Removes and returns the latest account (account with the largest account index).
+    pub async fn remove_latest_account(&mut self) -> crate::Result<Option<AccountHandle>> {
```
Thoralf-M: I think this should return (), because if you still have the accounthandle and call a method on it which would store the account again, it wouldn't be removed

samuel-rufi: good point!

Thoralf-M:
Please also expose it in the message interface

src/account_manager/mod.rs
```diff
@@ -76,6 +76,45 @@ impl AccountManager {
...
+                    largest_account_index_opt = Some(account_index);
+                }
+            } else {
+                largest_account_index_opt = Some(account_index)
+            }
+        }
+
+        if let Some(largest_account_index) = largest_account_index_opt {
+            let mut i = 0;
+            while i < accounts.len() {
```
Thoralf-M: Wouldn't `for i in 0..accounts.len(){}` be better?

samuel-rufi: much better

src/account_manager/mod.rs
```diff
@@ -76,6 +76,43 @@ impl AccountManager {
         Ok(self.accounts.read().await.clone())
     }
 
+    /// Removes the latest account (account with the largest account index).
+    pub async fn remove_latest_account(&mut self) -> crate::Result<()> {
```
Thoralf-M: I think this shouldn't be `mut`, causes only issues in the message interface and I think is also not needed, since the accounts are in `Arc<RwLock<>>`

samuel-rufi: it's needed because of the `pub async fn remove_latest_account(&mut self) -> crate::Result<()>`  in `AccountManager` if I'm not missing something?

samuel-rufi: But it is bit annoying because we have only this function that needs `mut`

Thoralf-M: What I say is that it should be changed exactly there

samuel-rufi: Adding mut to account_manager/mod.rs is not enough, also the bindings glue code would need to be updated

Thoralf-M: Make it `pub async fn remove_latest_account(&self) -> crate::Result<()> {`

samuel-rufi: oh wait i guess this mut is not required

samuel-rufi: removed it, wondering why the compiler didn't mention it :carlos:

src/message_interface/message_handler.rs
```diff
@@ -94,7 +94,7 @@ impl WalletMessageHandler {
     }
 
     /// Handles a message.
-    pub async fn handle(&self, message: Message, response_tx: UnboundedSender<Response>) {
+    pub async fn handle(&mut self, message: Message, response_tx: UnboundedSender<Response>) {
```
Thoralf-M:
```suggestion
    pub async fn handle(&self, message: Message, response_tx: UnboundedSender<Response>) {
```

src/message_interface/message_handler.rs
```diff
@@ -267,7 +267,11 @@ impl WalletMessageHandler {
         Ok(Response::Ok(()))
     }
 
-    async fn call_account_method(&self, account_id: &AccountIdentifier, method: &AccountMethod) -> Result<Response> {
+    async fn call_account_method(
+        &mut self,
```
Thoralf-M: This also needs to be undone

src/message_interface/mod.rs
```diff
@@ -71,7 +71,7 @@ pub async fn create_message_handler(options: Option<ManagerOptions>) -> crate::R
     Ok(WalletMessageHandler::with_manager(manager))
 }
 
-pub async fn send_message(handle: &WalletMessageHandler, message: Message) -> Response {
+pub async fn send_message(handle: &mut WalletMessageHandler, message: Message) -> Response {
```
Thoralf-M: Same here

tests/accounts.rs
```diff
@@ -41,6 +41,104 @@ async fn account_ordering() -> Result<()> {
...
+
+        let third_account = manager.create_account().finish().await?;
+        let fourth_account = manager.create_account().finish().await?;
+
+        assert_eq!(manager.get_accounts().await.unwrap().len(), 2);
+
+        let third_account_index = *third_account.read().await.index();
+        let fourth_account_index = *fourth_account.read().await.index();
+
+        (third_account_index, fourth_account_index)
```
Thoralf-M: The names are a little bit confusing, because the account indexes would be 0 and 1
Also it might be better to only recreate one account, so when we restore it from the db, one account is still removed, if re add both accounts before restoring it, the db might not be updated and the check wouldn't notice it, since we recreated the accounts already

samuel-rufi: Good point, changing it

samuel-rufi: When recreating only 1 account, the test fails now. Do you have an idea why this happens?
After restoring it finds 2 accounts, but it should be only 1

samuel-rufi: Seems like that the database does not get updated in-time. But this shouldn't happen, right? 
The drops should only happen once the database was updated successfully.

Thoralf-M (approved):

PR #1152, opened 2022-06-09
Use NativeTokens struct in AccountBalance

### LINKED ISSUE #1140: [Task]: Add up remaining balance fields
### Task description

Currently when the balance method from the account manager is called, not all values are accumulated
https://github.com/iotaledger/wallet.rs/blob/2e23b87da0aee2dfcdd27183b358bba686a7a536/src/account_manager/mod.rs#L192

Below in the sync() it's happening already, so this could be used as reference

### Requirements

N/A

### Acceptance criteria

N/A

### Creation checklist

oddgrd:
# Description of change

- Refactor `AccountBalance` to use `bee_block::output::NativeTokens`, and `Vec<NativeToken>` for `AccountBalanceDto`.
- Update nodejs `AccountBalance` type.
- Fill out the `balance` method on `AccountManager`.

Fixes issue #1140.

## Type of change

Ran examples to verify output was as expected.

## Change checklist

### DIFF
--- a/bindings/nodejs/types/account.ts
+++ b/bindings/nodejs/types/account.ts
@@ -1,5 +1,6 @@
+import type { INativeToken } from '@iota/types';
 import type { Address, AddressWithUnspentOutputs } from './address';
-import type { OutputData } from './output';
+import type { OutputData, OutputId } from './output';
 import type { Transaction } from './transaction';
 
 /**
@@ -12,20 +13,18 @@ export interface AccountBalance {
     total: string;
     available: string;
     requiredStorageDeposit: string;
-    nativeTokens: nativeTokensMap;
+    nativeTokens: INativeToken[];
     nfts: string[];
     aliases: string[];
     foundries: string[];
-    potentiallyLockedOutputs: potentiallyLockedOutputsMap;
+    potentiallyLockedOutputs: potentiallyLockedOutputs;
 }
 
-export type nativeTokensMap = {
-    [tokenId: string]: number;
-};
-
-export type potentiallyLockedOutputsMap = {
-    [outputId: string]: boolean;
-};
+/**
+ * Outputs with multiple unlock conditions and if they can currently be spent or not. If there is a
+ * TimelockUnlockCondition or ExpirationUnlockCondition this can change at any time
+ */
+export type potentiallyLockedOutputs = [OutputId, boolean];
 
 export interface AccountSyncOptions {
     addresses?: string[];
--- a/bindings/nodejs/types/account.ts
+++ b/bindings/nodejs/types/account.ts
@@ -36,7 +36,7 @@ export interface AccountSyncOptions {
      */
     addressStartIndex?: number;
     /**
-     * Usually we skip syncing if it's called within a few seconds, because there can only be new changes every 10
+     * Usually we skip syncing if it's called within a few seconds, because there can only be new changes every 5
      * seconds. But if we change the client options, we need to resync, because the new node could be from a nother
      * network and then we need to check all addresses. This will also ignore `address_start_index` and sync all
      * addresses. Default: false.
--- a/bindings/nodejs/types/account.ts
+++ b/bindings/nodejs/types/account.ts
@@ -1,6 +1,6 @@
 import type { INativeToken } from '@iota/types';
 import type { Address, AddressWithUnspentOutputs } from './address';
-import type { OutputData, OutputId } from './output';
+import type { OutputData } from './output';
 import type { Transaction } from './transaction';
 
 /**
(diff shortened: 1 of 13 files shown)


bindings/nodejs/types/output.ts
```diff
@@ -4,6 +4,11 @@ import type {
     IOutputMetadataResponse,
 } from '@iota/types';
 
+/**
+ * The identifier of an Output
+ */
+export type OutputId = string;
```
Tuditi: Is it worth defining a OutputId as a new string type? I feel like it's a bit of an overkill, since we could just use string everywhere.

oddgrd: It's just a personal preference, with a better comment it gives an opportunity to convey a bit more information than that it is a string. It is a string, but it's different from for example `blockId` which is also a string. Implementing this for OutputId means it should also be implemented for the other ID types though. I did this for `BlockId` in [iota.rs](https://github.com/iotaledger/iota.rs/blob/develop/bindings/nodejs/types/blockId.ts), but not the other types, so I should go back there and reconsider adding the other id types or removing `BlockId`. 

It's just an alias and doesn't do anything to verify that what is passed in is a valid OutputId. That can be done in typescript, but it is cumbersome and not without caveats, so I don't think that should be implemented. 

I don't feel strongly about this either way though, so let me know and I'll happily revert it. Thanks!

Tuditi: Thanks for your explanation (and your help with this!),  it makes sense. Let's keep it like that then :)!

bindings/nodejs/types/account.ts
```diff
@@ -12,30 +13,39 @@ export interface AccountBalance {
...
+     */
     addresses?: string[];
+    /**
+     * Address index from which to start syncing addresses. 0 by default, using a higher index will be faster because
+     * addresses with a lower index will be skipped, but could result in a wrong balance for that reason
+     */
     addressStartIndex?: number;
-    automaticOutputConsolidation?: boolean;
+    /**
+     * Usually we skip syncing if it's called within a few seconds, because there can only be new changes every 10
```
Thoralf-M: Milestone interval changed to 5 seconds (the comment in Rust code also needs to be updated)

oddgrd: I will fix both.

Thoralf-M (dismissed):

Thoralf-M (dismissed):

bindings/nodejs/types/account.ts
```diff
@@ -12,30 +13,39 @@ export interface AccountBalance {
...
-    potentiallyLockedOutputs: potentiallyLockedOutputsMap;
+    /**
+     * Outputs with multiple unlock conditions and if they can currently be spent or not. If there is a
+     * TimelockUnlockCondition or ExpirationUnlockCondition this can change at any time
+     */
+    potentiallyLockedOutputs: Record<string, boolean>;
 }
 
-export type nativeTokensMap = {
-    [tokenId: string]: number;
```
Tuditi, line 23: I had some time now to read up on this. `Record` and index signatures are fully equivalent. `Record` being `Record<string, number> = { [P in string]: number }`.

However with Record, we now lose the information about the actual meaning of the keys. An example: Devs would not know that inside Record<string, number>, the string represents the `tokenId`. So after coming to this conclusion, I think it's better to undo the changes. What are your thoughts on this?

Tuditi: Maybe this should not be done for nativeTokens, since the list approach makes sense (and there is no loss of information). But it might be better to revert it for potentiallyLockedOuputs. If we do that, we should rename the type to `PotentiallyLockedOuputs`, instead of `potentiallyLockedOutputsMap`, since it's not a real map.

bindings/nodejs/types/account.ts
```diff
@@ -12,30 +13,39 @@ export interface AccountBalance {
...
-    automaticOutputConsolidation?: boolean;
+    /**
+     * Usually we skip syncing if it's called within a few seconds, because there can only be new changes every 5
+     * seconds. But if we change the client options, we need to resync, because the new node could be from a nother
+     * network and then we need to check all addresses. This will also ignore `address_start_index` and sync all
+     * addresses. Default: false.
+     */
     forceSyncing?: boolean;
+    /** Checks pending transactions and promotes/reattaches them if necessary.  Default: true. */
     syncPendingTransactions?: boolean;
```
Tuditi, line 46:
```suggestion
    syncTransactions?: boolean;
```

I noticed an inconsistency on the rust side. An alternative is adjusting serde to `syncPendingTransactions` here: https://github.com/iotaledger/wallet.rs/blob/develop/src/account/operations/syncing/options.rs#L24

Tuditi (changes requested):
Thanks a lot for your work :rocket: ! I have two minor issues from the bindings side. :D

bindings/nodejs/types/account.ts
oddgrd: I agree, I think the named key with the indexed object notation is nice. The name of the key also shows up in the compiled types, unlike a type aliased primitive.

Tuditi (approved):
LGTM! :rocket:

Thoralf-M (approved):

PR #1229, opened 2022-06-29
Fix melting Native Token with foundry metadata

### LINKED ISSUE #1226: [Bug]: Native token melting fails if foundry has immutable metadata
### Issue description

Noticed through cli-wallet, but the bug is in wallet.rs:

The following method implements native token melting:
https://github.com/iotaledger/wallet.rs/blob/7bddeadaf6165285999c816116155083312cf5a9/src/account/operations/transaction/high_level/burning_melting/melt_native_token.rs#L40-L67

The problem in the method is on L53-L67, namely the creation of the next foundry output from the consumed one. The code does not account for immutable features in the consumed foundry (metadata for example), therefore the foundry transition and hence the transaction is invalid because it removes immutable features. (InvalidChainTransition, code 12).

Bonus question: What happens on L42?
https://github.com/iotaledger/wallet.rs/blob/7bddeadaf6165285999c816116155083312cf5a9/src/account/operations/transaction/high_level/burning_melting/melt_native_token.rs#L42
Why are we adding the balance of the consumed foundry to the alias? What's the point? The storage deposit requirement of the foundry will not change if you increase melted tokens count, so I'd suggest just use the balance of the consumed foundry for the output side foundry.

### Version

develop

### Expected behaviour

Melting doesn't fail if foundry has immutable features.

### Actual behaviour

It does.

### Can the issue reliably be reproduced?

Yes

### Steps to reproduce the issue

1. mint native tokens with metadata in the foundry
2. try to melt some

### Errors

_No response_

### Duplicate declaration

- [X] I have searched the issues tracker this issue and there is none

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

thibault-martinez:
Fixes #1226 

## Change checklist

### DIFF
--- a/src/account/operations/transaction/high_level/burning_melting/melt_native_token.rs
+++ b/src/account/operations/transaction/high_level/burning_melting/melt_native_token.rs
@@ -17,7 +17,6 @@ impl AccountHandle {
         options: Option<TransactionOptions>,
     ) -> crate::Result<TransactionResult> {
         log::debug!("[TRANSACTION] melt_native_token");
-        let byte_cost_config = self.client.get_byte_cost_config().await?;
 
         let token_id = native_token.0;
         let melt_token_amount = native_token.1;
@@ -44,7 +43,6 @@ impl AccountHandle {
             let outputs = vec![
                 alias_output,
                 FoundryOutputBuilder::from(&existing_foundry_output)
-                    .with_minimum_storage_deposit(byte_cost_config)
                     .with_token_scheme(TokenScheme::Simple(SimpleTokenScheme::new(
                         *token_scheme.minted_tokens(),
                         token_scheme.melted_tokens() + melt_token_amount,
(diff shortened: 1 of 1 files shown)


src/account/operations/transaction/high_level/burning_melting/melt_native_token.rs
```diff
@@ -38,33 +34,23 @@ impl AccountHandle {
...
-                        )?),
-                    )?
-                    .add_unlock_condition(UnlockCondition::ImmutableAliasAddress(
-                        ImmutableAliasAddressUnlockCondition::new(AliasAddress::from(alias_id)),
-                    ))
-                    .finish()?,
-                ),
+                alias_output,
+                FoundryOutputBuilder::from(&existing_foundry_output)
+                    .with_minimum_storage_deposit(byte_cost_config)
```
Thoralf-M: That should also not be required

Thoralf-M (approved):

samuel-rufi (approved):

PR #1253, opened 2022-07-04
Node.js Ledger Nano support

Thoralf-M:
# Description of change

Node.js Ledger Nano support

## Type of change

## Change checklist

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -22,6 +22,9 @@
           "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[1]' events"
         },
         {
+          "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[2]' ledger_nano"
+        },
+        {
           "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[2]' storage"
         },
         {
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -25,13 +25,13 @@
           "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[2]' ledger_nano"
         },
         {
-          "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[2]' storage"
+          "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[3]' storage"
         },
         {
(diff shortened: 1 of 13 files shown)


.changes/config.json
```diff
@@ -21,6 +21,9 @@
         {
           "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[1]' events"
         },
+        {
+          "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[2]' ledger_nano"
+        },
         {
           "command": "dasel put string -f Cargo.toml '.dependencies.iota-wallet.features.[2]' storage"
```
Tuditi: Indices have to be updated correctly

bindings/nodejs/lib/AccountManager.ts
```diff
@@ -133,6 +133,13 @@ export class AccountManager {
         return JSON.parse(response).payload;
     }
 
+    async getLedgerStatus(): Promise<NodeInfoWrapper> {
```
Tuditi: Why do we return a `NodeInfoWrapper` here?

bindings/nodejs/types/secretManager.ts
```diff
@@ -1,3 +1,7 @@
+export interface LedgerNanoSecretManager{}
+
+export interface LedgerNanoSimulatorLedgerNanoSecretManager{}
```
Tuditi: Doesn't really make sense to do this, since both are the same interface (empty object). Furthermore empty interfaces are not good design practice, since they work differently than in Rust. Maybe an option would be to provide a wrapper interface that contains a boolean for `isSimulator` property?

Tuditi: As an alternative, you could define a `Ledger` enum with two possible values (`Simulator/Hardware`).

Tuditi (changes requested):

bindings/nodejs/lib/AccountManager.ts
Thoralf-M: Because it's wrong :trollface:

bindings/nodejs/types/secretManager.ts
Thoralf-M: Opened a PR in iota.rs to remove the simulator variant and add the `isSimulator` bool

Tuditi (approved):
Looks good to me!

thibault-martinez (approved):

Issue #1221, opened 2022-06-28
Add and expose a new method to get a foundry output by native token id


Thoralf-M:
## Description

Get a foundry output by a native token id and return it

## Motivation

Currently one just sees the native token amount, but doesn't know more about them
It's expected that many native tokens will be minted with a native token metadata standard which will be put in the foundry, for that it would be useful to get the foundry by the native token id

## Requirements

1. Request foundry output by native token id

## Are you planning to do it yourself in a pull request?

Maybe.

PR #1256, opened 2022-07-04
Get foundry output by native token ID

oddgrd:
# Description of change

- Implement a method on `AccountHandle` to get `FoundryOutput` by `TokenId`: `get_foundry_output`. It first tries to get the foundry output from the account, and if it isn't in the account it will try to get it from the node
- Expose the new method in the message interface.
- Replace `BuiltOutput` `Response` variant with `OutputDto` and use it for the new method.
- Add examples to nodejs bindings: `22-mint-native-tokens` and `23-get-foundry-output`.
- Fix incorrect type in nodejs `NativeTokenOptions`.

Fixes issue #1221.

## Type of change

- Breaking change

I added an example in the nodejs bindings to mint native tokens, and then I tested the new method with various `TokenId`s, verifying that the returned foundry output is the same that minted the `NativeToken`.

## Change checklist

### DIFF
--- a/src/account/handle.rs
+++ b/src/account/handle.rs
@@ -5,7 +5,7 @@ use std::{ops::Deref, sync::Arc};
 
 use iota_client::{
     bee_block::{
-        output::OutputId,
+        output::{FoundryId, Output, OutputId, TokenId},
         payload::transaction::{TransactionId, TransactionPayload},
     },
     bee_rest_api::types::responses::OutputResponse,
@@ -74,6 +74,26 @@ impl AccountHandle {
     pub async fn get_output(&self, output_id: &OutputId) -> Option<OutputData> {
         let account = self.read().await;
         account.outputs().get(output_id).cloned()
+    }
+
+    /// Get the [`Output`] that minted a native token by the token ID. First try to get it
+    /// from the account, if it isn't in the account try to get it from the node
+    pub async fn get_foundry_output(&self, native_token_id: TokenId) -> Result<Output> {
+        let account = self.read().await;
+        let foundry_id = FoundryId::from(native_token_id);
+        for output_data in account.outputs().values() {
+            if let Output::Foundry(foundry_output) = &output_data.output {
+                if foundry_output.id() == foundry_id {
+                    return Ok(output_data.output.clone());
+                }
+            }
+        }
+
+        // Foundry was not found in the account, try to get it from the node
+        let foundry_output_id = self.client.foundry_output_id(foundry_id).await?;
+        let output_response = self.client.get_output(&foundry_output_id).await?;
+
+        Ok(Output::try_from(&output_response.output)?)
     }
 
     /// Get the [`Transaction`] of a transaction stored in the account
--- a/src/account/handle.rs
+++ b/src/account/handle.rs
@@ -88,6 +88,7 @@ impl AccountHandle {
                 }
             }
         }
+        drop(account);
 
         // Foundry was not found in the account, try to get it from the node
         let foundry_output_id = self.client.foundry_output_id(foundry_id).await?;
--- a/src/message_interface/response.rs
+++ b/src/message_interface/response.rs
@@ -51,7 +51,7 @@ pub enum Response {
     /// [`BuildNftOutput`](crate::message_interface::AccountMethod::BuildNftOutput)
     /// [`GetFoundryOutput`](crate::message_interface::AccountMethod::GetFoundryOutput)
     /// [`PrepareOutput`](crate::message_interface::AccountMethod::PrepareOutput)
-    OutputDto(OutputDto),
+    Output(OutputDto),
     /// Response for
     /// [`MinimumRequiredStorageDeposit`](crate::message_interface::AccountMethod::MinimumRequiredStorageDeposit)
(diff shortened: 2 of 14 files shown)


src/account/handle.rs
```diff
@@ -76,6 +76,26 @@ impl AccountHandle {
         account.outputs().get(output_id).cloned()
     }
 
+    /// Get the [`OutputData`] of a foundry output by native token id. First try to get it
```
Thoralf-M: OutputData is outdated

src/message_interface/response.rs
```diff
@@ -57,6 +57,8 @@ pub enum Response {
     /// [`GetOutputsWithAdditionalUnlockConditions`](crate::message_interface::AccountMethod::
     /// GetOutputsWithAdditionalUnlockConditions)
     OutputIds(Vec<OutputId>),
+    /// Response for [`GetFoundryOutput`](crate::message_interface::AccountMethod::GetFoundryOutput)
+    FoundryOutput(OutputDto),
```
Thoralf-M: @Tuditi would be another breaking change, but what do you think about merging `BuiltOutput(OutputDto),` and `FoundryOutput(OutputDto),` and just name it `OutputDto(OutputDto),`? Or would you rather want to keep them separated?

Tuditi: Is this a breaking change, due to the interface change of `NativeTokenOptions`?

As far as I can see this would work. If `FoundryOutput` would contain the signature, blockId, transactionId of the `alias` address that created it, I would keep them separate. Why are these not included in the response of `GetFoundryOutput`?

I couldn't test the response, due to `minting failed Alias output creation took too long`.

Thoralf-M: Why are you interested in the alias address and other stuff? Don't you just want to get the metadata?

Tuditi: It could make sense if we want to display the token creation date? Not sure if this is necessary, but I just wanted to clarify it for myself as well. 

That being said, I think your suggestion makes sense, since the objects are exactly the same!

Thoralf-M: Well I don't know what you really need :trollface:
So far the discussion was only about the native token metadata, which will be in the foundry output if the creator follows the standard

Tuditi: All good then, I agree that we should merge the objects.

Thoralf-M: @oddgrd we discussed a little bit more in private, please merge the responses to a single one and use it

oddgrd: Alright, will do!

src/account/handle.rs
```diff
@@ -76,6 +76,26 @@ impl AccountHandle {
...
+        let account = self.read().await;
+        let foundry_id = FoundryId::from(native_token_id);
+        for output_data in account.outputs().values() {
+            if let Output::Foundry(foundry_output) = &output_data.output {
+                if foundry_output.id() == foundry_id {
+                    return Ok(output_data.output.clone());
+                }
+            }
+        }
+
```
Thoralf-M, line 92: I think you should drop the account here before the client requests, so the lock is not held so long when not required

oddgrd: Good point, I will do that.

Thoralf-M (dismissed):

samuel-rufi (dismissed):

src/message_interface/response.rs
```diff
@@ -49,7 +49,9 @@ pub enum Response {
     /// [`BuildBasicOutput`](crate::message_interface::AccountMethod::BuildBasicOutput)
     /// [`BuildFoundryOutput`](crate::message_interface::AccountMethod::BuildFoundryOutput)
     /// [`BuildNftOutput`](crate::message_interface::AccountMethod::BuildNftOutput)
-    BuiltOutput(OutputDto),
+    /// [`GetFoundryOutput`](crate::message_interface::AccountMethod::GetFoundryOutput)
+    /// [`PrepareOutput`](crate::message_interface::AccountMethod::PrepareOutput)
+    OutputDto(OutputDto),
```
thibault-martinez: For consistency, I am not sure we should have `Dto` in the name, see `Account(AccountDto)` for example. I understand `Output` is already taken by `Output(Option<Box<OutputDataDto>>)` but maybe this one should be `OutputData(Option<Box<OutputDataDto>>)` instead? What do you think @Thoralf-M ?

Thoralf-M: Agree, this would actually be the better way

oddgrd: I agree as well, I will implement it. Sidenote: do you think `Outputs(Vec<OutputDataDto>)` should also be renamed then? It's a bit tricky since `OutputData` is already plural.

thibault-martinez: Hmm yeah that's tricky. `OutputsData` or `OutputsDatas` ? Idk 😬

oddgrd: I think `OutputsData` is okay. I can't think of anything better at least 😄

oddgrd: I went with that, happy to change if someone has a better idea. Thanks for the review!

Thoralf-M (approved):

samuel-rufi (approved):

PR #1278, opened 2022-07-08
Fix set_client_options to include coin type

Thoralf-M:
# Description of change

Fix set_client_options to include coin type
Also removed the syncing part, because the default sync options might not be wanted, if the node is from the same network there shouldn't be a change anyways and the user can sync himself afterwards

Fixes https://github.com/iotaledger/cli-wallet/issues/124

## Type of change

With changing the node multiple times in the cli wallet

## Change checklist

### DIFF
--- a/src/account_manager/builder.rs
+++ b/src/account_manager/builder.rs
@@ -4,7 +4,7 @@
 #[cfg(any(feature = "storage", feature = "stronghold"))]
 use std::path::PathBuf;
 use std::sync::{
-    atomic::{AtomicU32, AtomicUsize},
+    atomic::{AtomicU32, AtomicUsize, Ordering},
     Arc,
 };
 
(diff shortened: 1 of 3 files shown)


src/account_manager/builder.rs
```diff
@@ -220,4 +220,13 @@ impl AccountManagerBuilder {
             storage_manager,
         })
     }
+
+    pub(crate) async fn from_account_manager(account_manager: &AccountManager) -> Self {
```
thibault-martinez, line 224: Maybe just impl `From`? Bit more idiomatic

Thoralf-M: Doesn't work since it needs to be async

thibault-martinez: Ah right my bad overlooked that

thibault-martinez (approved):

Issue #1117, opened 2022-06-02
[Task]: Fix output collection


Thoralf-M:
### Task description

Outputs are collected in batches of 5 in collect_outputs_internal, at the end of each batch the storage deposit is evaluated. One source for funds of these storage deposits are the 'possible_additional_inputs' provided to the function.
If the number of outputs exceeds the batch size of 5 the code will not take into account whether or not the 'possible_additional_inputs' have already been used and might attempt to use the same 'possible_additional_inputs' multiple times.
https://github.com/iotaledger/wallet.rs/blob/6d28c47649130c06636c4431cabdd759d39afe05/src/account/operations/output_collection.rs#L316-L337

Outputs are collected in batches of 5 in collect_outputs_internal, at the end of each batch the storage deposit is evaluated. If the funds in the current outputs is unsuffient to meet the storage requirements additional inputs are considered.
The code mentions (correctly) that the storage costs needs to be updated each iteration of this loop, since the newly considered inputs might also produce new native tokens which will increase the storage requirements.
However the re-calculation uses the same input data in each iteration of the loop and therefore will return the same result each time it is called. This is caused because 'option_native_token' is a clone of new_native_tokens and only 'new_native_tokens' is updated in the loop.
https://github.com/iotaledger/wallet.rs/blob/6d28c47649130c06636c4431cabdd759d39afe05/src/account/operations/output_collection.rs#L319-L335

### Requirements

- Prevent reusing already used inputs when creating multiple consolidation transactions
- Re-calculate the input data for the increased storage deposit requirements

### Acceptance criteria

N/A

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

PR #1293, opened 2022-07-13
Fix claim outputs

samuel-rufi:
Fixes https://github.com/iotaledger/wallet.rs/issues/1117


src/account/operations/output_claiming.rs
```diff
@@ -225,93 +230,82 @@ impl AccountHandle {
...
+                            if let Some(native_tokens) = output_data.output.native_tokens() {
+                                new_native_tokens.add_native_tokens(native_tokens.clone())?;
+                                // Re-calculate the required storage deposit amount since we have added a new native
+                                // token
+                                required_storage_deposit = minimum_storage_deposit(
+                                    &byte_cost_config,
+                                    &first_account_address.address.inner,
+                                    &option_native_token,
+                                )?;
+                            }
```
Thoralf-M: Hm we do that in 288 already each time before we compare, so I guess this is not needed

src/account/operations/output_claiming.rs
```diff
@@ -225,93 +230,82 @@ impl AccountHandle {
...
-            let mut additional_inputs = Vec::new();
 
+            let mut additional_inputs = Vec::new();
             if new_amount < required_storage_deposit {
                 // add more inputs
                 for output_data in &possible_additional_inputs {
                     // Recalculate every time, because new intputs can also add more native tokens, which would increase
                     // the storage deposit cost
-                    let required_storage_deposit = minimum_storage_deposit(
+                    let mut required_storage_deposit = minimum_storage_deposit(
```
Thoralf-M: But maybe here it should be like
```suggestion
                    required_storage_deposit = minimum_storage_deposit(
```
and the first declaration should be mutable

samuel-rufi: fixed

src/account/operations/output_claiming.rs
samuel-rufi: fixed

Thoralf-M (dismissed):

Thoralf-M (approved):

thibault-martinez (approved):

PR #1295, opened 2022-07-13
Add `note` field

### LINKED ISSUE #1060: Add optional note to transactions
## Description

Add an optional note `Option<String>` to the `Transaction` and `TransferOptions` structs which will be stored locally

## Motivation

For some use cases it's useful to be able to set some data related to a transaction which will be stored in the db and also saved in backups
For example for Firefly with the current participation structure it's not possible to know the action (started participating for an event or stopped participating) just by looking at the transaction, adding the optional note would make it easily possible

## Requirements

1. Add  `Option<String>` to `Transaction` and `TransferOptions`
2. When `TransferOptions` has it set, add it to the created transaction

Maybe.

pvdrz:


src/account/types/mod.rs
```diff
@@ -208,6 +208,8 @@ pub struct Transaction {
     pub network_id: u64,
     // set if the transaction was created by the wallet or if it was sent by someone else and is incoming
     pub incoming: bool,
+
+    pub note: Option<String>,
```
Thoralf-M, line 211: This field also needs to be added to the TransactionDto below

Thoralf-M (dismissed):

src/account/types/mod.rs
```diff
@@ -208,6 +208,8 @@ pub struct Transaction {
     pub network_id: u64,
     // set if the transaction was created by the wallet or if it was sent by someone else and is incoming
     pub incoming: bool,
+
```
thibault-martinez:
```suggestion
```

src/account/types/mod.rs
```diff
@@ -228,6 +229,8 @@ pub struct TransactionDto {
     pub network_id: String,
     /// If the transaction was created by the wallet or if it was sent by someone else and is incoming
     pub incoming: bool,
+
```
thibault-martinez:
```suggestion
```

thibault-martinez (approved):

Thoralf-M (approved):

PR #1297, opened 2022-07-14
Increase consolidation inputs

Thoralf-M:
# Description of change

Increase max consolidation inputs amount

## Type of change

Modified example

## Change checklist

### DIFF
--- a/src/account/constants.rs
+++ b/src/account/constants.rs
@@ -3,9 +3,9 @@
 
 /// Amount at which outputs on a single addresses will get consolidated by default if consolidatioin is enabled
 pub(crate) const DEFAULT_OUTPUT_CONSOLIDATION_THRESHOLD: usize = 100;
-#[cfg(feature = "ledger_nano")]
 /// Amount at which outputs on a single addresses will get consolidated by default with a ledger secret_manager if
 /// consolidatioin is enabled, needs to be smaller because the memory of the ledger nano s is limited
+#[cfg(feature = "ledger_nano")]
 pub(crate) const DEFAULT_LEDGER_OUTPUT_CONSOLIDATION_THRESHOLD: usize = 15;
 
 /// Amount of API request that can be sent in parallel during syncing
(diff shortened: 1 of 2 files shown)


src/account/constants.rs
```diff
@@ -6,7 +6,7 @@ pub(crate) const DEFAULT_OUTPUT_CONSOLIDATION_THRESHOLD: usize = 100;
 #[cfg(feature = "ledger_nano")]
```
thibault-martinez: Put the attribute after the comment

src/account/operations/output_consolidation.rs
```diff
@@ -1,13 +1,24 @@
...
+    },
 };
 #[cfg(feature = "ledger_nano")]
 use iota_client::secret::SecretManager;
 
+// Constants for the calculation of the amount of inputs we can use with a ledger nano
+#[cfg(feature = "ledger_nano")]
+const ESSENCE_SIZE_WITHOUT_IN_AND_OUTPUTS: usize = 49;
+#[cfg(feature = "ledger_nano")]
+const INPUT_SIZE_IN_ESSENCE: usize = 43;
```
thibault-martinez: How do you get 43 ?

src/account/operations/output_consolidation.rs
```diff
@@ -1,13 +1,24 @@
...
+#[cfg(feature = "ledger_nano")]
+const MIN_OUTPUT_SIZE_IN_ESSENCE: usize = 46;
```
thibault-martinez, line 21: Same question, I get 45 but I'm probably missing something

src/account/operations/output_consolidation.rs
Thoralf-M: Got the number from Thomas
```Rust
    let input = Input::Utxo(UtxoInput::from_str("0xb66fd384cb5755668f1890ea2e41d699db9cf32f3bc422ad3c24ffeb9c7f01d00000").unwrap());
    println!("{}", input.packed_len());
```
prints 35, so I guess this is the correct number to use?

Thoralf-M: Got the number and din't think further about it :trollface: 
It's 35 + 8 for the bip32 path part we need in addition to it, will add a comment and change the name

src/account/operations/output_consolidation.rs
Thoralf-M: I get 45 for a basic output, but then we need to also add the output type byte

src/account/operations/output_consolidation.rs
thibault-martinez: Ah I see, wasn't accounting for the bip32 path, thanks

thibault-martinez (approved):

PR #1343, opened 2022-08-08
Sync addresses from confirmed transaction outputs

### LINKED ISSUE #1189: Add outputs from confirmed transactions to unspent outputs
## Description

Currently after sending a transaction we also call retry_until_included on it which would check if it gets confirmed
If it gets confirmed then, we could already add the remainder outputs to the account again and make them available
We would need to check if the account addresses contain the address of the address unlock condition, expiration unlock conditions should be ignored I think.

## Motivation

That way one can send transactions out without syncing as long as funds are still available

## Requirements

1. Add outputs from confirmed to unspent outputs without syncing
2. When syncing update the outputs with the correct metadata

## Open questions (optional)

Should we modify retry_until_included to return the metadata also?

Probably.

Thoralf-M:
# Description of change

Sync addresses from confirmed transaction outputs, so the outputs get available without the need of manually syncing
Also handle some output stuff in syncing, so we don't make outputs spent, even though they aren't, just because we don't sync them

Fixes #1189 

## Type of change

Running the threads example modified so it doesn't call sync()

## Change checklist

### DIFF
--- a/src/account/operations/transaction/mod.rs
+++ b/src/account/operations/transaction/mod.rs
@@ -215,15 +215,14 @@ impl AccountHandle {
                                 .outputs()
                                 .iter()
                                 .filter_map(|o| {
-                                    o.unlock_conditions()
-                                        .expect("output needs to have unlock conditions")
-                                        .address()
-                                        .and_then(|output_address| {
+                                    o.unlock_conditions().and_then(|unlock_conditions| {
+                                        unlock_conditions.address().and_then(|output_address| {
                                             account_addresses
                                                 .iter()
                                                 .find(|a| a.address.inner == *output_address.address())
                                                 .map(|acc_address| acc_address.address.to_bech32())
                                         })
+                                    })
                                 })
                                 .collect();
 
--- a/src/account/operations/transaction/mod.rs
+++ b/src/account/operations/transaction/mod.rs
@@ -14,7 +14,15 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use iota_client::{
     api::{verify_semantic, PreparedTransactionData, SignedTransactionData},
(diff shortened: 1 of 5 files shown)


src/account/operations/transaction/mod.rs
```diff
@@ -178,4 +190,63 @@ impl AccountHandle {
         }
         Ok(())
     }
+
+    // Try to get a transaction confirmed and sync related account addresses when confirmed, so the outputs get
+    // available for new transactions without manually syncing (which would sync all addresses and be more heavy without
+    // extra logic)
+    fn monitor_tx_confirmation(&self, block_id: BlockId) {
+        // spawn a thread which tries to get the block confirmed
```
thibault-martinez:
```suggestion
        // spawn a task which tries to get the block confirmed
```

src/account/operations/transaction/mod.rs
```diff
@@ -178,4 +190,63 @@ impl AccountHandle {
...
+                    if let Some(Payload::Transaction(tx_payload)) = confirmed_block.1.payload() {
+                        let TransactionEssence::Regular(regular_essence) = tx_payload.essence();
+                        if let Ok(account_addresses) = account.list_addresses().await {
+                            // Filter for addresses from the account
+                            let addresses: Vec<String> = regular_essence
+                                .outputs()
+                                .iter()
+                                .filter_map(|o| {
+                                    o.unlock_conditions()
+                                        .expect("output needs to have unlock conditions")
```
thibault-martinez: I know this can't happen atm but it's probably more future-proof to also use an `and_then` here

thibault-martinez (approved):

samuel-rufi (approved):

PR #1359, opened 2022-08-15
Use iota-crypto aead helpers

thibault-martinez:
## Change checklist


src/storage/manager.rs
```diff
@@ -178,7 +179,7 @@ fn parse_accounts(accounts: &[String], encryption_key: &Option<[u8; 32]>) -> cra
         let account_json = if account.starts_with('{') {
             Some(account.to_string())
         } else if let Some(key) = encryption_key {
-            Some(decrypt_record(account, key)?)
+            Some(String::from_utf8(chacha::aead_decrypt(key, account.as_bytes())?).unwrap())
```
Thoralf-M: Should this really be `from_utf8`? We had `from_utf8_lossy` before and maybe users would want to use not valid utf8 as `note` for a transaction and then it would fail? Not really sure, but probably good to check and if we want to limit it to utf8 we should probably validate it there and also check if there could be other places where it could cause problems

thibault-martinez: Yeah this is why I didn't merge this one yet. I just quickly made it so it compiles but I need to think about it a bit more 👍🏻

thibault-martinez: Changed it

Thoralf-M (approved):

PR #1390, opened 2022-08-24
Update rocksdb

Thoralf-M:


thibault-martinez:
Doesn't seem to contain anything clippy related or... ?

Thoralf-M:
> Doesn't seem to contain anything clippy related or... ?

Yeah not anymore after I rebased

thibault-martinez:
> > Doesn't seem to contain anything clippy related or... ?
> 
> Yeah not anymore after I rebased

But clippy is still failing :trollface:

src/storage/mod.rs
```diff
@@ -542,6 +542,7 @@ fn split_key_checksum(record: &[u8]) -> crate::Result<([u8; XChaCha20Poly1305::N
     Ok((nonce, kcv))
 }
 
+#[allow(clippy::let_and_return)]
```
Thoralf-M: Can't just remove the `let is_valid = ` because of 
```
error[E0597]: `storage_handle` does not live long enough
   --> src/storage/mod.rs:559:19
    |
559 |             match storage_handle.lock().await.get(ACCOUNT_INDEXATION_KEY).await {
    |                   ^^^^^^^^^^^^^^^^^^^^^------
    |                   |
    |                   borrowed value does not live long enough
    |                   a temporary with access to the borrow is created here ...
...
576 |         }
    |         -
    |         |
    |         `storage_handle` dropped here while still borrowed
    |         ... and the borrow might be used here, when that temporary is dropped and runs the `Drop` code for type `tokio::sync::MutexGuard`
    |
    = note: the temporary is part of an expression at the end of a block;
            consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
    |
559 ~             let x = match storage_handle.lock().await.get(ACCOUNT_INDEXATION_KEY).await {
560 |                 // Existing DB
  ...
574 |                 Err(e) => Err(e),
575 ~             }; x
    |
```

thibault-martinez: You could technically combine `let storage_handle = crate::storage::get(storage_path).await?;` and `storage_handle.lock().await.get(ACCOUNT_INDEXATION_KEY).await` into one assignment to get rid of the warning

thibault-martinez (approved):

PR #1411, opened 2022-09-06
Check max native tokens limit in consolidation and claiming

### LINKED ISSUE #1353: Invalid native token count
image
caused when syncing
I think this is caused by https://github.com/iotaledger/wallet.rs/blob/ca0c7680d99ac40bbfcc4146644cb4816a4a749b/src/account/operations/balance.rs#L34, because this type has the validation for the native tokens in a single output

Thoralf-M:
# Description of change

Check max native tokens limit in consolidation and claiming

Fixes https://github.com/iotaledger/wallet.rs/issues/1353

## Type of change

Running consolidate and claim command in the cli wallet with outputs that have native tokens

## Change checklist

### DIFF
--- a/src/account/operations/output_claiming.rs
+++ b/src/account/operations/output_claiming.rs
@@ -233,7 +233,7 @@ impl AccountHandle {
             for output_data in outputs {
                 if let Some(native_tokens) = output_data.output.native_tokens() {
                     // Skip output if the max native tokens count would be exceeded
-                    if new_native_tokens.len() + native_tokens.len() > NativeTokens::COUNT_MAX.into() {
+                    if get_new_native_token_count(&new_native_tokens, native_tokens)? > NativeTokens::COUNT_MAX.into() {
                         log::debug!("[OUTPUT_CLAIMING] skipping output to not exceed the max native tokens count");
                         continue;
                     }
@@ -391,4 +391,16 @@ pub(crate) fn sdr_not_expired(output: &Output, current_time: u32) -> Option<&Sto
     } else {
         None
     }
+}
+
+// Helper function to calculate the native token count without duplicates, when new native tokens are added
+// Might be possible to refactor the sections where it's used to remove the clones
+pub(crate) fn get_new_native_token_count(
+    native_tokens_builder: &NativeTokensBuilder,
+    native_tokens: &NativeTokens,
+) -> crate::Result<usize> {
+    // Clone to get the new native token count without actually modifying it
+    let mut native_tokens_count = native_tokens_builder.clone();
+    native_tokens_count.add_native_tokens(native_tokens.clone())?;
+    Ok(native_tokens_count.len())
 }
--- a/src/account/operations/output_claiming.rs
+++ b/src/account/operations/output_claiming.rs
@@ -9,7 +9,7 @@ use iota_client::{
         address::Address,
         output::{
             unlock_condition::{AddressUnlockCondition, StorageDepositReturnUnlockCondition, UnlockCondition},
-            BasicOutputBuilder, NativeTokensBuilder, NftOutputBuilder, Output, OutputId,
+            BasicOutputBuilder, NativeTokens, NativeTokensBuilder, NftOutputBuilder, Output, OutputId,
         },
     },
 };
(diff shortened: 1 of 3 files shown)


samuel-rufi (dismissed):

src/account/operations/output_claiming.rs
```diff
@@ -231,6 +231,24 @@ impl AccountHandle {
             let mut new_native_tokens = NativeTokensBuilder::new();
             // check native tokens
             for output_data in outputs {
+                if let Some(native_tokens) = output_data.output.native_tokens() {
+                    // Skip output if the max native tokens count would be exceeded
+                    if new_native_tokens.len() + native_tokens.len() > NativeTokens::COUNT_MAX.into() {
+                        log::debug!("[OUTPUT_CLAIMING] skipping output to not exceed the max native tokens count");
+                        continue;
+                    }
+                    new_native_tokens.add_native_tokens(native_tokens.clone())?;
```
thibault-martinez, line 240: A side-node: we could avoid that kind of clones by taking ownership of `outputs_to_claim` since it's not used anymore after. But it seems like the std `chunks` doesn't work with ownership and we would have to use the `itertools` crate. Not a terribly bad idea if we can remove some nested clones but maybe something that can be done later and not in this PR.

src/account/operations/output_claiming.rs
```diff
@@ -231,6 +231,24 @@ impl AccountHandle {
             let mut new_native_tokens = NativeTokensBuilder::new();
             // check native tokens
             for output_data in outputs {
+                if let Some(native_tokens) = output_data.output.native_tokens() {
+                    // Skip output if the max native tokens count would be exceeded
+                    if new_native_tokens.len() + native_tokens.len() > NativeTokens::COUNT_MAX.into() {
```
thibault-martinez: Is this correct? Because at this point there could be same Token IDs in both sides and they would be counted twice so we could ignore an output that was actually fine.

thibault-martinez: It's claiming anyway, not a very big deal if we are being overly cautious I guess

src/account/operations/output_claiming.rs
```diff
@@ -291,10 +296,17 @@ impl AccountHandle {
                     )?;
                     if new_amount < required_storage_deposit {
                         if !additional_inputs_used.contains(&output_data.output_id) {
-                            new_amount += output_data.output.amount();
                             if let Some(native_tokens) = output_data.output.native_tokens() {
+                                // Skip input if the max native tokens count would be exceeded
+                                if new_native_tokens.len() + native_tokens.len() > NativeTokens::COUNT_MAX.into() {
```
thibault-martinez: Same question as above

src/account/operations/output_claiming.rs
Thoralf-M: Should I do the following? It's two additional clones
```Rust
                    // Clone to get the new native token count without actually modifying it
                    let mut native_tokens_count = new_native_tokens.clone();
                    native_tokens_count.add_native_tokens(native_tokens.clone())?;
                    // Skip output if the max native tokens count would be exceeded
                    if native_tokens_count.len() > NativeTokens::COUNT_MAX.into() {
```

src/account/operations/output_consolidation.rs
```diff
@@ -124,13 +123,19 @@ impl AccountHandle {
                 let mut custom_inputs = Vec::with_capacity(max_inputs.into());
                 let mut total_native_tokens = NativeTokensBuilder::new();
                 for output_data in outputs {
-                    total_amount += output_data.output.amount();
-
-                    custom_inputs.push(output_data.output_id);
-
                     if let Some(native_tokens) = output_data.output.native_tokens() {
+                        // Skip output if the max native tokens count would be exceeded
+                        if total_native_tokens.len() + native_tokens.len() > NativeTokens::COUNT_MAX.into() {
```
thibault-martinez: same

src/account/operations/output_claiming.rs
thibault-martinez: Not using it here?

src/account/operations/output_claiming.rs
Thoralf-M: Already thought about using itertools at another time for something, but didn't want to add it as dependency just because of that, if we have more things like this it probably makes sense

src/account/operations/output_claiming.rs
Thoralf-M: Forgot it, updated

src/account/operations/output_claiming.rs
thibault-martinez: I think it's generally not bad importing a new crate if it ends up costing less at runtime. We can do that at refactoring later.

thibault-martinez (dismissed):

thibault-martinez (commented):
Maybe add a .changes actually?

thibault-martinez (approved):

samuel-rufi (approved):

PR #1412, opened 2022-09-06
Add features missing from NftOptions

### LINKED ISSUE #1292: [Task]: Add NFTs to inputs
### Task description

NFTs that have more amount than needed for storage deposit should also be available in inputs to send for sending amounts. But the NFT shouldn't get burned, so if they're used, they should be added in the output side again.

Currently they are ignored here if we don't have them on the output side already https://github.com/iotaledger/wallet.rs/blob/f1acbe4d1484eaec25d87352f15ca7c04f13cbf5/src/account/operations/transaction/input_selection.rs#L153

If there is for example an nft output with 100 SMR, but storage deposit only needs to be 1, then the other 99 should still be available for transactions. We should then add the NFT output in the output side, just with the minimum storage deposit and the other amount can be used and remainder can go in a basic output as it's happening already now

The best way to handle this, might be to add the logic in iota.rs https://github.com/iotaledger/iota.rs/blob/76460deeba707f83c41142081e6b1cbaf1e9809a/src/api/block_builder/input_selection/mod.rs#L41

### Requirements

-

### Acceptance criteria

-

### Creation checklist

### LINKED ISSUE #1331: Features missing from NftOptions
## Description

- In `NftOptions` there is only metadata Features, but no way to add "Issuer Immutable Feature" (also no way to add "Sender Feature")

We should add support for those.

## Motivation

To support all fields possible in the Nft output

## Open questions (optional)

Does this make sense ?

No.

thibault-martinez:
Fixes #1331 

## Change checklist


Thoralf-M (commented):
Have you tried to create a transaction with the issuer and sender features? I'm not sure if it works right now or requires changes from the input selection

thibault-martinez:
> Have you tried to create a transaction with the issuer and sender features? I'm not sure if it works right now or requires changes from the input selection

Not yet, doing it now.

Thoralf-M:
https://github.com/thibault-martinez/wallet.rs/pull/2

Thoralf-M (dismissed):

samuel-rufi (dismissed):

Thoralf-M (approved):

Thoralf-M (changes requested):
NFTs are currently also added when not needed (might be fixed when we update client rev)

Update: fixed now

Thoralf-M (dismissed):

Thoralf-M (approved):

samuel-rufi (approved):

PR #1428, opened 2022-09-12
Make account meta private

### LINKED ISSUE #1391: Incoming output isn't synced on account after first login
## Bug description

In Firefly:

i am logged into an account and request some fund from the faucet to simulate a soon incoming output for my account. Then i log out and stay logged out until i'm sure the funds arrived (i checked the explorer). When i log back in, the account doesn't contain this output, only if i logout and again login, it appears

## Rust version

Which version of Rust are you running?

- Rust version: 2.0.2-alpha.21

## Steps To reproduce the bug

1. get output while offline
2. log into account
3. output isnt synced yet
4. logout
5. login
6. output is now there

## Expected behaviour

Output should be available after first login

## Actual behaviour

Output is available after second login

Thoralf-M:
# Description of change

Make account meta private, because the data there isn't updated, one should call the account methods instead to get up to date data

Fixes https://github.com/iotaledger/wallet.rs/issues/1391

## Type of change

## Change checklist


MarkNerdi996 (changes requested):
I just checked all the getter methods, and the methd `listIncomingTransactions()` is missing.  Could you please add that

src/account/handle.rs
```diff
@@ -188,7 +188,15 @@ impl AccountHandle {
         Ok(outputs)
     }
 
-    /// Returns all transaction of the account
+    /// Returns all incoming transactions of the account
+    pub async fn list_incoming_transactions(
+        &self,
+    ) -> Result<Vec<(TransactionId, (TransactionPayload, Vec<OutputResponse>))>> {
```
thibault-martinez: Just curious why a second tuple in the tuple?

thibault-martinez: Make it look like a Hashmap?

thibault-martinez (dismissed):

src/account/handle.rs
Thoralf-M: Hm yeah because it's one, probably better to also return it as one, just remembered that there was something with returning it in nodejs as hashmap, but that's also fine
Will change

thibault-martinez (dismissed):

samuel-rufi (dismissed):

thibault-martinez (changes requested):
Conflicts

thibault-martinez (changes requested):
Conflict :trollface:

bindings/nodejs/lib/Account.ts
```diff
@@ -216,6 +220,34 @@ export class Account {
         return JSON.parse(resp).payload;
     }
 
+    /**
+     * Melt native tokens. This happens with the foundry output which minted them, by increasing it's
```
thibault-martinez:
```suggestion
     * Melt native tokens. This happens with the foundry output which minted them, by increasing its
```

thibault-martinez (approved):

MarkNerdi996 (approved):
Looks good for me! 
If a differentiation between `publicAddresses` and `internalAddresses` is needed, i'll let you know

PR #1433, opened 2022-09-14
Update recover accounts

### LINKED ISSUE #1425: Add an `accountStartIndex` to `AccountSyncOptions`
## Description

Balance finding algorithms could be improved if we can specify a starting index for the accounts in addition to the addresses. If we can only adjust the starting address index then it's possible that any new accounts searched will not start from address index zero, which may result in funds not found.

Another improvement consists of using the gap limit `j` as follows:

0) Set address range to equal the gap limit  (`range = [0, j]`)
1) Look through the address range for addresses with funds . 
2) Two scenarios are posibble:
 - If no address is found go to step 3. 
 - If an address is found at `i < j`. Then update the address range to `[i, i + j]` and go to Step 1.
3) Stop searching

## Motivation

See description.

## Requirements

Write a list of what you want this feature to do.

1. Allow for a `accountStartIndex` parameter in the `AccountSyncOptions`
2. Allow both the `accountStartIndex` and `addressStartIndex` to override the start in the case where the start index is lower than where funds were actually found, e.g. if wallet-rs knows funds reside on address 10 but we specify to start at address 5 then it should start at address 5

Maybe (@Tuditi and I would like to try and do this if not too difficult).

Thoralf-M:
# Description of change

Update recover accounts

Fixes https://github.com/iotaledger/wallet.rs/issues/1425

## Type of change

Running recover accounts example

## Change checklist

### DIFF
--- a/src/account/operations/output_finder.rs
+++ b/src/account/operations/output_finder.rs
@@ -8,37 +8,55 @@ use iota_client::secret::GenerateAddressMetadata;
 use crate::account::{
     handle::AccountHandle,
     operations::{address_generation::AddressGenerationOptions, syncing::SyncOptions},
+    types::AddressWithUnspentOutputs,
 };
 
 impl AccountHandle {
     /// Search addresses with unspent outputs
-    /// `address_gap_limit` defines how many addresses without outputs will be checked in each account, if an address
-    /// has outputs, the counter is reset
+    /// `address_gap_limit`: The number of addresses to search for, after the last address with unspent outputs
     /// Addresses that got crated during this operation and have a higher key_index than the latest one with outputs,
     /// will be removed again, to keep the account size smaller
     pub(crate) async fn search_addresses_with_outputs(
         self: &AccountHandle,
-        address_gap_limit: u32,
+        mut address_gap_limit: u32,
         mut sync_options: Option<SyncOptions>,
     ) -> crate::Result<usize> {
         log::debug!("[search_addresses_with_outputs]");
 
-        // store the length so we can remove addresses with higher indexes later if they don't have outputs
+        let mut address_gap_limit_internal = address_gap_limit;
+
+        // store the current index, so we can remove new addresses with higher indexes later again, if they don't have
+        // outputs
         let (highest_public_address_index, highest_internal_address_index) = {
             let account = self.read().await;
-            let highest_public_address_index = match account.public_addresses.last() {
-                Some(a) => a.key_index,
-                None => 0,
-            };
-            let highest_internal_address_index = match account.internal_addresses.last() {
-                Some(a) => a.key_index,
-                None => 0,
-            };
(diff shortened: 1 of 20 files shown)


samuel-rufi (dismissed):

src/message_interface/message.rs
```diff
@@ -78,17 +78,18 @@ pub enum Message {
...
         #[serde(rename = "accountGapLimit")]
-        /// Defines how many accounts without unspent outputs will be
-        /// checked, if an account has unspent outputs, the counter is reset
+        /// The number of accounts to search for, after the last account with unspent outputs.
         account_gap_limit: u32,
         #[serde(rename = "addressGapLimit")]
-        /// Defines how many addresses without unspent outputs will be checked in each account, if an
-        /// address has unspent outputs, the counter is reset
+        /// The number of addresses to search for, after the last address with unspent outputs, in
+        ///   each account.
```
thibault-martinez:
```suggestion
        /// each account.
```

src/message_interface/message.rs
```diff
@@ -78,17 +78,18 @@ pub enum Message {
...
-        /// Defines how many addresses without unspent outputs will be checked in each account, if an
-        /// address has unspent outputs, the counter is reset
+        /// The number of addresses to search for, after the last address with unspent outputs, in
+        /// each account.
         address_gap_limit: u32,
-        /// SyncOptions to be used during the account recovery process.
-        /// address_start_index and force_syncing will be overwritten in sync_options to not skip addresses, but also
-        /// don't send duplicated requests
+        /// Optional parameter to specify the sync options. The `address_start_index` and `force_syncing`
+        ///   fields will be overwritten to skip existing addresses.
```
thibault-martinez:
```suggestion
        /// fields will be overwritten to skip existing addresses.
```

src/account/operations/output_finder.rs
```diff
@@ -8,37 +8,55 @@ use iota_client::secret::GenerateAddressMetadata;
...
+            // outside
+            let (highest_public_address_index, highest_internal_address_index) = {
+                let account = self.read().await;
+                (
+                    account
+                        .public_addresses
+                        .last()
+                        .map(|a| a.key_index)
+                        .expect("account needs to have a public address"),
+                    account.internal_addresses.last().map(|a| a.key_index),
```
thibault-martinez, line 54: Is there no other way at all to do that? Because to full iterations inside an iteration is not idea

Thoralf-M: What do you mean with full iteration?

thibault-martinez: Oh this is `Vec::last` and not `Iterator::last`, ignore me 🙈

thibault-martinez (approved):

samuel-rufi (approved):

PR #1434, opened 2022-09-14
Remove try_claim_outputs()

### LINKED ISSUE #1417: 1 function 1 transaction
## Description

Don't have functions that could create multiple transactions internally

## Motivation

Better error handling, clearer logic

## Requirements

1. Change claiming and consolidation to only send one transaction at a time, where the user has to provide the output ids (return error if they don't fit in a single tx)
2. Change minting to return an error if there is no existing alias output, add a new method instead to create an alias output
3. Change burn destroy methods to return an error if there is still an output owned by the output we're about to burn/destroy? That way the users would need to burn/destroy the outputs at the end of the ownership chain first (this should also make it harder to burn/destroy something by accident)

## Open questions (optional)

Any strong reasons to not do this? For claiming and consolidation it might require the users to create a loop to consolidate/claim all outputs, but that shouldn't be hard to do

Yes/No.

Thoralf-M:
# Description of change

Remove try_claim_outputs() and only do a single claim tx when calling the claim function

Partially fixes #1417 

## Type of change

cli wallet

## Change checklist

### DIFF
--- a/src/message_interface/message_handler.rs
+++ b/src/message_interface/message_handler.rs
@@ -800,9 +800,7 @@ impl WalletMessageHandler {
             AccountMethod::ClaimOutputs { output_ids_to_claim } => {
                 convert_async_panics(|| async {
                     let transactions = account_handle.claim_outputs(output_ids_to_claim.to_vec()).await?;
-                    Ok(Response::SentTransactions(
-                        transactions.iter().map(TransactionDto::from).collect(),
-                    ))
+                    Ok(Response::SentTransaction(TransactionDto::from(&transactions)))
                 })
                 .await
             }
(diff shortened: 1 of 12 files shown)


samuel-rufi (dismissed):

src/message_interface/message_handler.rs
```diff
@@ -797,21 +797,10 @@ impl WalletMessageHandler {
...
-                    let transactions = account_handle.try_claim_outputs(outputs_to_claim).await?;
-                    Ok(Response::SentTransactions(
-                        transactions.iter().map(TransactionDto::from).collect(),
-                    ))
-                })
-                .await
-            }
             AccountMethod::ClaimOutputs { output_ids_to_claim } => {
                 convert_async_panics(|| async {
                     let transactions = account_handle.claim_outputs(output_ids_to_claim.to_vec()).await?;
```
thibault-martinez: Singular ?

thibault-martinez (approved):

samuel-rufi (approved):

PR #1449, opened 2022-09-21
Fix spent outputs

Thoralf-M:
# Description of change

Fix wrongly spent outputs. When we sync but don't get output ids from the index we set them as spent, even though they could still be unspent, but we just didn't get them anymore because of changed query parameters when syncing.
Since we want to update the metadata, we request it and based on that we can check if it's really spent or not. This was just not checked before.

## Type of change

Cli wallet

## Change checklist


src/account/update.rs
```diff
@@ -40,7 +40,7 @@ impl AccountHandle {
         &self,
         addresses_with_unspent_outputs: Vec<AddressWithUnspentOutputs>,
         unspent_outputs: Vec<OutputData>,
-        spent_outputs: Vec<OutputId>,
+        spent_or_not_synced_output_ids: Vec<OutputId>,
         spent_or_not_synced_output_responses: Vec<OutputResponse>,
```
thibault-martinez: Are these two fields 1:1 related ? If yes, any way to have it be a HashMap or too annoying from the calling site? Would avoid the double loop.

thibault-martinez: Do we even need the ids at all ? Seems to me we could achieve the same without it at all

Thoralf-M: We need the ids because if the output is pruned and the node can't return the metadata, then we can't set it as spent if we just work with the metadata we get

thibault-martinez: Can't we imagine anything like a HashMap of Option OutputResponse or whatever? Fine if we can't, but that double loop is a bit annoying

Thoralf-M: Yes HashMap with Option should work

samuel-rufi (dismissed):

src/account/update.rs
```diff
@@ -100,7 +99,23 @@ impl AccountHandle {
...
 
         // Update spent outputs
-        for output_id in spent_outputs {
+        for (output_id, output_response) in spent_or_not_synced_outputs {
+            // If we got the output response and it's still unspent, skip it
+            if let Some(output_response) = output_response {
+                if output_response.metadata.is_spent {
+                    // Update output_response if it got spent to include the new metadata
+                    let transaction_id = TransactionId::from_str(&output_response.metadata.transaction_id)?;
+                    let output_id = OutputId::new(transaction_id, output_response.metadata.output_index)?;
```
thibault-martinez, line 108: Don't you already have the output_id ?

thibault-martinez: You already have the output_id?

.changes/tx monitoring.md
```diff
@@ -0,0 +1,5 @@
+---
```
thibault-martinez: Space in the file name :trollface: ?

Thoralf-M: Why not? :trollface:

thibault-martinez: Because it triggers my brain :trollface:

thibault-martinez (approved):

samuel-rufi (approved):

PR #1485, opened 2022-10-05
Add danger note on top of the guide

samuel-rufi:


documentation/docs/how_tos/exchange_guide.mdx
```diff
@@ -27,9 +27,22 @@ You can easily integrate wallet.rs with your exchange, custody solution, or prod
...
-This guide explains how to implement the Wallet library in your exchange.
+:::warning
+
+IOTA allows for complex tokenization schemes, but they can lead to monetary losses if used improperly or integrated incorrectly.
+Transaction outputs may have multiple unlocking conditions which may require returning some or all of the amount, which could expire if not claimed in time, or which may not be unlockable for a very long time.
+Therefore, pay attention when working with transactions, understand what kind of unlock conditions exist and test your integration properly.
+
+Examples:
+- if the transaction produces an output that can be unlocked using an "expiration unlock condition", the sender could claim the money back if the recipient did not claim it in time
+- if the transaction produces an output that can be unlocked using an "timelock unlocking condition", it might be the case that receiver can claim the funds never or only after a long time
```
Thoralf-M: This is a lot of text already, I feel like it's already too much at this point
Maybe we should have a shorter warning like `You can loose funds, if you don't check the unlock conditions before accepting a deposit!` which then links to another page which has the longer explanation (then we can also link to that one from other places)
We could also add the storage deposit return unlock condition in the example and also example code how to check these things to accept only most basic outputs and be safe (unlock condition length == 1, unlock condition type is address unlock condition, address is an ed25519 address)

What do you think?

samuel-rufi: @thibault-martinez did want have details :trollface:

samuel-rufi: But your sentence brings it to point too I guess

Thoralf-M: Yeah sure we still want to have details, I just thing all of them in the warning will just be a big red block with much text and then they will just skip reading it :carlos:

samuel-rufi: I agree, people might skip it when its too long

thibault-martinez: Yeah I just meant we can't say "you can lose money" without saying how it can actually happen.

documentation/docs/how_tos/exchange_guide.mdx
```diff
@@ -27,9 +27,22 @@ You can easily integrate wallet.rs with your exchange, custody solution, or prod
 
 :::
 
-## Implementation Guide
 
-This guide explains how to implement the Wallet library in your exchange.
+:::warning
+
+IOTA allows for complex tokenization schemes, but they can lead to monetary losses if used improperly or integrated incorrectly.
```
thibault-martinez: Shimmer ?

Thoralf-M (approved):

thibault-martinez (approved):

PR #1501, opened 2022-10-18
Add Java publish workflow

samuel-rufi:


.github/workflows/java_binding_publish.yml
```diff
@@ -0,0 +1,95 @@
+name: Publish Java binding
+
+on:
+  push:
+    branches:
+      - staging
```
Thoralf-M: :carloshuh:

samuel-rufi: what should i replace it with? :carlos:

Thoralf-M: develop? We have no staging branch

Thoralf-M: But should probably also not run on every push

Thoralf-M: Ok maybe we will get staging :trollface:

PR #1508, opened 2022-10-20
Test engine

thibault-martinez:
## Change checklist


tester/src/transactions.rs
```diff
@@ -0,0 +1,92 @@
...
+                }
+            } else {
+                return Err(Error::MissingField("outputs"));
+            }
+
+            let _transaction = context.account_manager.get_accounts().await?[0]
+                .send(outputs, None)
+                .await?;
+
+            time::sleep(Duration::from_secs(10)).await;
```
Thoralf-M: If that's for waiting that the tx gets confirmed, it might be better to use 
```Rust
    if let Some(block_id) = transaction.block_id{
        account
        .retry_until_included(&block_id, None, None)
        .await?;
    }else{
        account.sync(None).await?;
        time::sleep(Duration::from_secs(10)).await;
    }
```

thibault-martinez: Done, thanks

tester/json/consume_two_outputs.json
```diff
@@ -0,0 +1,39 @@
+{
+    "accounts": [
+        [
+            75000,
+            75000
+        ],
+        0
```
samuel-rufi, line 7: this 0 means that there exists some account 2 but has no balance yet ?

tester/json/consume_two_outputs.json
```diff
@@ -0,0 +1,39 @@
...
+    ],
+    "steps": [
+        {
+            "transaction": {
+                "from": 0,
```
samuel-rufi, line 12: does this describe the sender ? therefore account 0 ?

tester/json/consume_two_outputs.json
thibault-martinez: Yes

tester/json/consume_two_outputs.json
thibault-martinez: Yes

tester/src/checks/balance.rs
```diff
@@ -0,0 +1,43 @@
...
+use serde_json::Value;
+
+use crate::{context::Context, error::Error};
+
+pub async fn process_balance<'a>(context: &Context<'a>, balance: &Value) -> Result<(), Error> {
+    let account_index = if let Some(account_index) = balance.get("account") {
+        if let Some(account_index) = account_index.as_u64() {
+            account_index as usize
+        } else {
+            return Err(Error::InvalidField("account"));
```
Thoralf-M:
```suggestion
            return Err(Error::InvalidField("account_index"));
```

tester/src/checks/balance.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use serde_json::Value;
+
+use crate::{context::Context, error::Error};
+
+pub async fn process_balance<'a>(context: &Context<'a>, balance: &Value) -> Result<(), Error> {
+    let account_index = if let Some(account_index) = balance.get("account") {
+        if let Some(account_index) = account_index.as_u64() {
```
Thoralf-M, line 10: Max account index is u32::MAX, is there also an as_u32?

tester/src/main.rs
```diff
@@ -0,0 +1,136 @@
...
+        if let Err(err) = process_json(&context, json).await {
+            log::error!(
+                "Executing test {}/{}: {:?} failed: {}.",
+                index + 1,
+                paths.len(),
+                path.file_name().unwrap(),
+                err
+            );
+        }
+    }
```
Thoralf-M: After all tests, if there was an error, we need to add an exit code so the CI fails

tester/src/checks/balance.rs
thibault-martinez: No :(

tester/src/main.rs
thibault-martinez: Will do

.github/workflows/tester.yml
```diff
@@ -0,0 +1,89 @@
+name: Wallet tester
+
+on:
+  push:
+    branches:
+      - test-engine-ci
+      - dev
```
thibault-martinez:
```suggestion
      - develop
```

.github/workflows/tester.yml
```diff
@@ -0,0 +1,89 @@
...
+
+on:
+  push:
+    branches:
+      - test-engine-ci
+      - develop
+  pull_request:
+    branches:
+      - develop
+      - test-engine-ci
```
Thoralf-M:
```suggestion
      - develop
  pull_request:
    branches:
      - develop
```

tester/src/accounts/account.rs
```diff
@@ -0,0 +1,22 @@
...
+    error::Error,
+};
+
+pub async fn process_account<'a>(context: &Context<'a>, account: &Value) -> Result<(), Error> {
+    let created_account = context.account_manager.create_account().finish().await?;
+
+    if account.is_u64() {
+        process_amount(context, &created_account, account.as_u64().unwrap()).await?;
+    } else if account.is_array() {
+        process_outputs(context, &created_account, account.as_array().unwrap()).await?;
```
Thoralf-M: Why not just this one here? The extra function with another name is not really clear here
```suggestion
    for output in account.as_array().unwrap() {
        if output.is_u64() {
            process_amount(context, account, output.as_u64().unwrap()).await?;
        }
    }
```
Also the whole naming, why is it called `account` if it's the amount?

thibault-martinez: Because there will be more types of output in the future, amount is just one option to quickly create a basic output with this value

thibault-martinez: And same reason for the naming, account will not always be an amount

tester/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    let args: Vec<String> = std::env::args().collect();
+    let mut error = false;
+
+    let mnemonic = String::from("average day true meadow dawn pistol near vicious have ordinary sting fetch mobile month ladder explain tornado curious energy orange belt glue surge urban");
```
Thoralf-M, line 87:
```suggestion
    // private tangle faucet mnemonic: https://github.com/iotaledger/hornet/blob/develop/private_tangle/private_tangle_keys.md#faucet
    let mnemonic = String::from("average day true meadow dawn pistol near vicious have ordinary sting fetch mobile month ladder explain tornado curious energy orange belt glue surge urban");
```

tester/src/accounts/account.rs
Thoralf-M: But then let's call it `output` or why `account`?

thibault-martinez: Because it's an array of outputs

Thoralf-M: Ok then why not call it `outputs`? :carlos:

thibault-martinez: I'm afraid that if we call the field `outputs`, it's not easy to understand that it's actually creating an account per index

thibault-martinez: And accounts is more flexible, we may want to set more things than outputs in the future

Thoralf-M: Huh, where is it creating a new account?

thibault-martinez: L13

Thoralf-M: Yes but that's before and only a single account independent of one or more

thibault-martinez: I can rename it but it's going to be obsolete very soon as we'll need to actually have an account object instead of what we have now

thibault-martinez: Changing anyway

thibault-martinez: Done :carlos:

Thoralf-M (approved):

samuel-rufi (approved):

Issue #1531, opened 2022-11-01
Implement `AccountManager` Governance methods


maxwellmattryan:
## Description

:no_entry_sign: Blocked until technical specification is solidified.

The base functionality of registering Governance proposals for a Firefly profile will be necessary as part of a main user flow.

See `AccountManager`'s [Methods](https://hackmd.io/@maxwellmattryan/r1zCOAaGs/%2FBH6LvsTYQWWjymIHQ061Pw#Methods) in the technical specification.

## Motivation

Required for Firefly's Governance feature.

## Requirements

Refer to the methods in the [technical specification](https://hackmd.io/@maxwellmattryan/r1zCOAaGs/%2FBH6LvsTYQWWjymIHQ061Pw#Methods).

## Open questions (optional)

N/a

## Are you planning to do it yourself in a pull request?

Some where I can.

PR #1590, opened 2022-11-17
Add remaining account manager methods to message interface

Thoralf-M:
# Description of change

Add remaining account manager methods to message interface

https://github.com/iotaledger/wallet.rs/issues/1531

## Type of change

## Change checklist

### DIFF
--- a/src/message_interface/message_handler.rs
+++ b/src/message_interface/message_handler.rs
@@ -318,7 +318,7 @@ impl WalletMessageHandler {
                         .account_manager
                         .register_participation_event(event_id, nodes)
                         .await?;
-                    Ok(Response::ParticipationEvent(event))
+                    Ok(Response::ParticipationEvent(Some(event)))
                 })
                 .await
             }
@@ -337,8 +337,7 @@ impl WalletMessageHandler {
                         .account_manager
                         .get_participation_event(event_id)
                         .await?
-                        .ok_or_else(|| crate::Error::Storage(format!("event {event_id} not found")))?
-                        .0;
+                        .map(|(e, _)| e);
                     Ok(Response::ParticipationEvent(event))
                 })
                 .await
--- a/src/message_interface/message_handler.rs
+++ b/src/message_interface/message_handler.rs
@@ -330,6 +330,35 @@ impl WalletMessageHandler {
                 })
                 .await
             }
+            #[cfg(feature = "participation")]
+            Message::GetParticipationEvent(event_id) => {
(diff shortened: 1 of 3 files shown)


src/message_interface/message_handler.rs
```diff
@@ -330,6 +330,35 @@ impl WalletMessageHandler {
...
                 .await
             }
+            #[cfg(feature = "participation")]
+            Message::GetParticipationEvent(event_id) => {
+                convert_async_panics(|| async {
+                    let event = self
+                        .account_manager
+                        .get_participation_event(event_id)
+                        .await?
+                        .ok_or_else(|| crate::Error::Storage(format!("event {event_id} not found")))?
```
thibault-martinez: Feels a bit odd to return an error, I would put an option in the Response, we do that for some other types

src/message_interface/message_handler.rs
```diff
@@ -330,6 +330,34 @@ impl WalletMessageHandler {
...
+                        .await?
+                        .map(|(e, _)| e);
+                    Ok(Response::ParticipationEvent(event))
+                })
+                .await
+            }
+            #[cfg(feature = "participation")]
+            Message::GetParticipationEventStatus(event_id) => {
+                convert_async_panics(|| async {
+                    let event_status = self.account_manager.get_participation_event_status(event_id).await?;
```
thibault-martinez, line 348: I know I previously approved this but I also think this should return Option and not an error if the record is not found

Thoralf-M: But this is the event status which gets requested from the node

thibault-martinez: Not sure what you mean. What happens if you pass a bullshit event ID?

Thoralf-M: Then it should return an error, that the event wasn't found and not None

thibault-martinez: Why does `get_participation_event` return an option and not `get_participation_event_status`? They are doing pretty similar stuff ?

Thoralf-M: get_participation_event works with local data from the db, get_participation_event_status returns what the node returns for this event

thibault-martinez: Ok I see. I now remember asking in the past why we don't return Option for the routes, this still kinda puzzles me a bit. But fine 👍🏻

thibault-martinez (approved):

samuel-rufi (approved):

PR #1602, opened 2022-11-22
Add InitializeWalletException

samuel-rufi:


bindings/java/iota-wallet-java/lib/native/src/lib.rs
```diff
@@ -47,6 +69,11 @@ pub extern "system" fn Java_org_iota_api_NativeApi_sendMessage(
     _class: JClass,
     command: JString,
 ) -> jstring {
+
+    if env.exception_check().unwrap() {
```
thibault-martinez, line 73: Safe to unwrap?

bindings/java/iota-wallet-java/lib/native/src/lib.rs
```diff
@@ -29,11 +29,33 @@ pub extern "system" fn Java_org_iota_api_NativeApi_createMessageHandler(
...
     config: JString,
 ) {
-    let manager_options: ManagerOptions = {
-        let input: String = env.get_string(config).expect("Couldn't get java string!").into();
-        serde_json::from_str(&input).unwrap()
+    let manager_options: ManagerOptions = match env.get_string(config) {
+        Ok(json_input) => match serde_json::from_str(&String::from(json_input)) {
+            Ok(manager_options) => manager_options,
+            Err(err) => {
+                env.throw_new("java/lang/Exception", err.to_string()).unwrap();
```
thibault-martinez, line 36: Are all these `throw_new` safe to unwrap?

bindings/java/iota-wallet-java/lib/native/src/lib.rs
samuel-rufi: I think this should be good.
Since if the JVM fails then failing in Rust at this point makes sense too I would say.

samuel-rufi: I mean how should we handle it differently? We would need to return something to the JVM (but if it fails in the first case then returning something to it doesn't make sense too).
So I guess panicing in Rust is fine at this point

bindings/java/iota-wallet-java/lib/native/src/lib.rs
samuel-rufi: Yes I would argue it the same way as for the above comment.
Maybe we should create an issue for it and let Brord have also a look, so we have another opinion

thibault-martinez: Yeah sure, do that please

samuel-rufi: oki

thibault-martinez (approved):

PR #1612, opened 2022-11-24
Split required storage deposit balance

### LINKED ISSUE #1574: Change balance
## Description

Remove required storage deposit of alias, nft and foundry outputs from the available balance, since it can't be used without burning them. Also remove the required storage deposit of basic outputs with native tokens.

Balance calculation is happening here https://github.com/iotaledger/wallet.rs/blob/f541e9c06f9c7f253c368c6562d47a49cc2acbc1/src/account/operations/balance.rs#L18

Also split the required storage deposit per output types like this:
```
required_storage_deposit: {
aliases: 235235,
basic_outputs: 13513537,
foundries: 2462467,
nfts: 3135135
}
```

## Motivation

Clearer to the user

No.

thibault-martinez:
Fixes #1574 

## Change checklist


samuel-rufi (dismissed):

src/account/operations/balance.rs
```diff
@@ -143,8 +173,19 @@ impl AccountHandle {
...
+                                if output_data.output.is_basic() {
+                                    let rent = output_data.output.rent_cost(&rent_structure);
+
+                                    required_storage_deposit.basic += rent;
+                                } else if output_data.output.is_nft() {
+                                    let rent = output_data.output.rent_cost(&rent_structure);
+
+                                    required_storage_deposit.nft += rent;
+                                    total_rent_amount += rent;
+                                }
```
Thoralf-M, line 187: How about changing this?
```suggestion
                                let rent = output_data.output.rent_cost(&rent_structure);
                                if output_data.output.is_basic() {
                                    required_storage_deposit.basic += rent;
                                    // amount for basic outputs isn't added to total_rent_amount, since we can also spent it without burning
                                } else if output_data.output.is_nft() {
                                    required_storage_deposit.nft += rent;
                                    total_rent_amount += rent;
                                }
```

Thoralf-M: How about changing this?
```suggestion
                                let rent = output_data.output.rent_cost(&rent_structure);
                                if output_data.output.is_basic() {
                                    required_storage_deposit.basic += rent;
                                    // amount for basic outputs isn't added to total_rent_amount, since we can also spend it without burning
                                } else if output_data.output.is_nft() {
                                    required_storage_deposit.nft += rent;
                                    total_rent_amount += rent;
                                }
```

src/account/operations/balance.rs
```diff
@@ -225,7 +270,8 @@ impl AccountHandle {
         Ok(AccountBalance {
             base_coin: BaseCoinBalance {
                 total: total_amount,
-                available: total_amount - locked_amount,
+                // TODO alles gut?
```
Thoralf-M: I don't think so
`locked_amount` is the full amount of the outputs in the inputs, but they are still in the unspent outputs, so the amount of these outputs that got added to `total_rent_amount` will be duplicated removed

Thoralf-M: Maybe we should just add something like this at the places where we add to the total amount
```Rust
if !locked_outputs.contains(output_id){ 
// add to total_rent_amount 
}
```

thibault-martinez: Should adding to `required_storage_deposit` also be in this if ?

Thoralf-M: Not sure, I would say rather not since we also show the amount of these outputs in the total balance

src/account/operations/balance.rs
```diff
@@ -190,11 +223,15 @@ impl AccountHandle {
...
             }
         }
+
         log::debug!(
             "[BALANCE] total_amount: {}, locked balance: {}",
             total_amount,
             locked_amount
         );
+
+        // TODO should we add total_rent_amount to locked_amount here ?
```
thibault-martinez: @Thoralf-M What about this ?

Thoralf-M: Yes I think that's what we said we want to do

thibault-martinez: Wait shouldn't it be a minus actually ? `total_amount < locked_amount - total_rent_amount`

Thoralf-M: My comment was about L 275 `available: total_amount - locked_amount - total_rent_amount,` where the minus is doing the same as adding it to the locked amount before
And in this check it needs to be + then, since we subtract both at the end, no?
Or just do `locked_amount += total_rent_amount;` before both things?

thibault-martinez: Yeah sounds good, did that

src/account/operations/balance.rs
```diff
@@ -143,8 +174,19 @@ impl AccountHandle {
 
                                 // Add amount
                                 total_amount += amount;
+
                                 // Add storage deposit
-                                required_storage_deposit += output_data.output.rent_cost(&rent_structure);
+                                if output_data.output.is_basic() {
+                                    required_storage_deposit.basic += rent;
+                                    // Amount for basic outputs isn't added to total_rent_amount, since we can spend it
+                                    // without burning.
```
Thoralf-M: Should we also add it if the output has native tokens as here? https://github.com/iotaledger/wallet.rs/pull/1612/files#diff-22aa72010988d33cf882f809863b6642515ef8df5cf339ed256377f5f7491194R106-R114

src/account/operations/balance.rs
```diff
@@ -143,8 +174,29 @@ impl AccountHandle {
...
                                 // Add amount
                                 total_amount += amount;
+
                                 // Add storage deposit
-                                required_storage_deposit += output_data.output.rent_cost(&rent_structure);
+                                if output_data.output.is_basic() {
+                                    required_storage_deposit.basic += rent;
+                                    // Amount for basic outputs isn't added to total_rent_amount if there aren't native
+                                    // tokens, since we can spend it
+                                    // without burning.
```
Thoralf-M:
```suggestion
                                    // tokens, since we can spend it without burning.
```

thibault-martinez: Weird, fmt did that

Thoralf-M (approved):

samuel-rufi (approved):

PR #1616, opened 2022-11-25
feat: add participation to nodejs bindings

Tuditi:
# Description of change

Make the participation methods of the account object callable through nodejs.

## Links to any relevant issues

closes: #1532 

## Type of change

Choose a type of change, and delete any options that are not relevant.

- Enhancement (a non-breaking change which adds functionality)

## How the change has been tested

Through example 28 by using a private tangle. The following curl adds an event to the private tangle:

```
curl -X POST http://localhost:14264/api/participation/v1/admin/events -H 'Content-Type: application/json' -d '{"name":"Tuditi Proposal","milestoneIndexCommence":100,"milestoneIndexStart":115,"milestoneIndexEnd":510700,"payload":{"type":0,"questions":[{"text":"Should Firefly release a token?","answers":[{"value":1,"text":"Yes","additionalInfo":""},{"value":2,"text":"No","additionalInfo":""}],"additionalInfo":""}]},"additionalInfo":"Nothing needed here"}'
```

## Change checklist

Tick the boxes that are relevant to your changes, and delete any items that are not.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

PR #1624, opened 2022-11-29
feat: add participation to account manager nodejs

Tuditi:
# Description of change

Merge #1616 first!

Add participation methods to account manager.

## Type of change

Added example 28 to test the participation methods on account manager. To run:
1. Publish an event to the Tangle (or private tangle)
2. Replace the constant `EVENT_ID` with your event id
3. Run `node 28-get-voting-information.js` from `bindings/nodejs/examples`

## Change checklist


thibault-martinez:
Can you rebase please 🙏🏻 ?

bindings/nodejs/types/participation.ts
```diff
@@ -0,0 +1,52 @@
...
+    questions?: Answer[]
+    checksum: string
+}
+
+interface EventData {
+    name: string
+    milestoneIndexCommence: number
+    milestoneIndexStart: number
+    milestoneIndexEnd: number
+    payload: VotingEventPayload
```
Thoralf-M: This could also be a staking payload, maybe you can copy the types from the mainnet Firefly version?

bindings/nodejs/Cargo.toml
```diff
@@ -18,7 +18,7 @@ exclude = [ "index.node" ]
 crate-type = [ "cdylib" ]
 
 [dependencies]
-iota-wallet = { path = "../../", default-features = false, features = [ "mnemonic", "events", "ledger_nano", "storage", "stronghold", "message_interface", "participation" ] }
+iota-wallet = { path = "../../", version = "1.0.0-rc.1", default-features = false, features = [ "mnemonic", "events", "ledger_nano", "storage", "stronghold", "message_interface", "participation" ] }
```
thibault-martinez: Why ?

bindings/nodejs/lib/AccountManager.ts
```diff
@@ -179,6 +192,33 @@ export class AccountManager {
...
+        const response = await this.messageHandler.sendMessage({
+            cmd: 'getParticipationEvent',
+            payload: {
+                eventId
+            }
+        })
+        return JSON.parse(response).payload
+    }
+
+    async getParticipationEventS(): Promise<Event[]> {
```
thibault-martinez: uppercase S ?

bindings/nodejs/Cargo.toml
Tuditi: Thanks for reviewing the PR, but it is still in draft mode, which is why it contains such things. Cargo added this when I was building the bindings, but I also think this is not needed, so I'll remove it.

thibault-martinez: Sorry, my bad. Will wait for your green light 🟢

bindings/nodejs/types/participation.ts
Tuditi: But do we want it to be a staking payload if we don't provide support for it? I don't think so, which is why at the moment, it's better to type it as `VotingEventPayload`. Once we want a staking payload, this should be updated.

Thoralf-M: But it can return staking payloads already, can't this break something for you if that's happening?

Tuditi: If we don't want staking events, how can it return a staking payload? Basically we are introducing extra code that we don't want to use anyway (for the foreseeable future). I suggest that we remove this from iota.rs (happy to do it myself as well): https://github.com/iotaledger/iota.rs/blob/91bd73383a95ee9de4bfba90c1bc9b922e8d4ca3/client/src/node_api/participation/types/mod.rs#L56. Maybe I missed a reason why we need to keep it, so enlighten me in that case pls :D

Thoralf-M: Someone can create a staking event on the participation plugin and then you get these from the node and it would return an error, because you only expect voting events

thibault-martinez (approved):

Thoralf-M (approved):

PR #1628, opened 2022-11-30
Remove voting output from inputs

Thoralf-M:
# Description of change

Remove voting output from inputs, so it doesn't get spent when not calling a function related to it

## Type of change

Examples having a voting output and then trying to mint an nft, the output doesn't get selected anymore

## Change checklist

### DIFF
--- a/src/account/operations/output_claiming.rs
+++ b/src/account/operations/output_claiming.rs
@@ -156,7 +156,7 @@ impl AccountHandle {
         for (output_id, output_data) in &account.unspent_outputs {
             #[cfg(feature = "participation")]
             if let Some(ref voting_output) = voting_output {
-                // Remove voting output from inputs, so it doesn't get spent when not calling a function related to it.
+                // Remove voting output from inputs, because we don't want to spent it to claim something else.
                 if output_data.output_id == voting_output.output_id {
                     continue;
                 }
--- a/src/account/operations/output_claiming.rs
+++ b/src/account/operations/output_claiming.rs
@@ -147,11 +147,20 @@ impl AccountHandle {
     /// additional inputs
     pub async fn get_basic_outputs_for_additional_inputs(&self) -> crate::Result<Vec<OutputData>> {
         log::debug!("[OUTPUT_CLAIMING] get_basic_outputs_for_additional_inputs");
+        #[cfg(feature = "participation")]
+        let voting_output = self.get_voting_output().await?;
         let account = self.read().await;
 
         // Get basic outputs only with AddressUnlockCondition and no other unlock condition
         let mut basic_outputs: Vec<OutputData> = Vec::new();
         for (output_id, output_data) in &account.unspent_outputs {
+            #[cfg(feature = "participation")]
+            if let Some(ref voting_output) = voting_output {
+                // Remove voting output from inputs, so it doesn't get spent when not calling a function related to it.
+                if output_data.output_id == voting_output.output_id {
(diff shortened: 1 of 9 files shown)


samuel-rufi (dismissed):

thibault-martinez (dismissed):
I think this would be better if we somehow remove the voting output from the available outputs instead of adding a very specific parameter to filter_inputs but approving since it'll all disappear with the new ISA.

samuel-rufi (dismissed):

src/account/operations/output_consolidation.rs
```diff
@@ -72,13 +72,22 @@ impl AccountHandle {
...
         let account = self.read().await;
         let account_addresses = &account.addresses_with_unspent_outputs[..];
         let current_time = self.client.get_time_checked().await?;
         let token_supply = self.client.get_token_supply().await?;
 
         let mut outputs_to_consolidate = Vec::new();
         for (output_id, output_data) in account.unspent_outputs() {
+            #[cfg(feature = "participation")]
+            if let Some(ref voting_output) = voting_output {
+                // Remove voting output from inputs, so it doesn't get spent when not calling a function related to it.
```
thibault-martinez: Comment could be a bit more specific to this case

src/account/operations/output_claiming.rs
```diff
@@ -147,11 +147,20 @@ impl AccountHandle {
...
+        #[cfg(feature = "participation")]
+        let voting_output = self.get_voting_output().await?;
         let account = self.read().await;
 
         // Get basic outputs only with AddressUnlockCondition and no other unlock condition
         let mut basic_outputs: Vec<OutputData> = Vec::new();
         for (output_id, output_data) in &account.unspent_outputs {
+            #[cfg(feature = "participation")]
+            if let Some(ref voting_output) = voting_output {
+                // Remove voting output from inputs, so it doesn't get spent when not calling a function related to it.
```
thibault-martinez: Same

thibault-martinez (dismissed):

src/account/operations/output_claiming.rs
Thoralf-M:
```suggestion
                // Remove voting output from inputs, because we don't want to spent it to claim something else.
```

src/account/operations/output_consolidation.rs
Thoralf-M:
```suggestion
                // Remove voting output from inputs, because we want to keep it's features and not consolidate it.
```

src/account/operations/output_consolidation.rs
```diff
@@ -72,13 +72,22 @@ impl AccountHandle {
...
         let account = self.read().await;
         let account_addresses = &account.addresses_with_unspent_outputs[..];
         let current_time = self.client.get_time_checked().await?;
         let token_supply = self.client.get_token_supply().await?;
 
         let mut outputs_to_consolidate = Vec::new();
         for (output_id, output_data) in account.unspent_outputs() {
+            #[cfg(feature = "participation")]
+            if let Some(ref voting_output) = voting_output {
+                // Remove voting output from inputs, because we want to keep it's features and not consolidate it.
```
thibault-martinez:
```suggestion
                // Remove voting output from inputs, because we want to keep its features and not consolidate it.
```

thibault-martinez (approved):

samuel-rufi (approved):

PR #1643, opened 2022-12-02
Add requestFundsFromFaucet() function

### LINKED ISSUE #1630: [Task]: Add Shimmer Java Code Examples #1588
## Description

Shimmer code examples using the Java binding to showcase the library's functionalities in all languages. 

## Motivation

As we aim to make developer onboarding easy in all languages, it'd be great to provide the same examples in all languages.

## Requirements

Add examples showcasing the following functionalities using Rust:

- [x] Request funds from the faucet (https://github.com/iotaledger/wallet.rs/pull/1643)

If no one's planning on working on these, I can take a go at it.

samuel-rufi:
Fixes https://github.com/iotaledger/wallet.rs/issues/1630


src/message_interface/account_method.rs
```diff
@@ -392,4 +392,10 @@ pub enum AccountMethod {
     /// Expected response: [`SentTransaction`](crate::message_interface::Response::SentTransaction)
     #[cfg(feature = "participation")]
     DecreaseVotingPower { amount: String },
+    /// Expected response: [`Faucet`](crate::message_interface::Response::Faucet)
+    RequestFundsFromFaucet {
+        #[serde(rename = "faucetUrl")]
+        faucet_url: String,
```
thibault-martinez: just name it url, it's obviously the faucet url

bindings/java/examples/src/RequestFundsFromFaucet.java
```diff
@@ -0,0 +1,41 @@
...
+import org.iota.types.*;
+import org.iota.types.account_methods.SyncAccount;
+import org.iota.types.exceptions.InitializeWalletException;
+import org.iota.types.exceptions.NoFundsReceivedFromFaucetException;
+import org.iota.types.exceptions.WalletException;
+import org.iota.types.ids.account.AccountAlias;
+import org.iota.types.secret.StrongholdSecretManager;
+
+public class RequestFundsFromFaucet {
+    public static void main(String[] args) throws WalletException, InterruptedException, InitializeWalletException, NoFundsReceivedFromFaucetException {
```
samuel-rufi, line 15:
```suggestion
    
    public static void main(String[] args) throws WalletException, InterruptedException, InitializeWalletException, NoFundsReceivedFromFaucetException {
```

bindings/java/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `requestFundsFromFaucet()` function and example;
```
thibault-martinez, line 26: Also put the `RequiredStorageDeposit` change (in Changed)

src/message_interface/account_method.rs
```diff
@@ -392,4 +392,9 @@ pub enum AccountMethod {
     /// Expected response: [`SentTransaction`](crate::message_interface::Response::SentTransaction)
     #[cfg(feature = "participation")]
     DecreaseVotingPower { amount: String },
+    /// Expected response: [`Faucet`](crate::message_interface::Response::Faucet)
+    RequestFundsFromFaucet {
```
thibault-martinez: Changelog

bindings/java/CHANGELOG.md
```diff
@@ -31,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Fix `RequiredStorageDeposit` deserialization;
```
thibault-martinez: If it's a fix, put it in Fixed

thibault-martinez (dismissed):

thibault-martinez (dismissed):

bindings/java/lib/src/main/java/org/iota/types/RequiredStorageDeposit.java
```diff
@@ -0,0 +1,28 @@
...
+package org.iota.types;
+
+public class RequiredStorageDeposit extends AbstractObject {
+
+    private String alias;
+    private String basic;
+    private String foundry;
+    private String nft;
+
+    public String getAlias() {
```
kwek20: shouldnt all these be `long`?

bindings/java/examples/src/RequestFundsFromFaucet.java
```diff
@@ -0,0 +1,42 @@
...
+        Wallet wallet = new Wallet(new WalletConfig()
+                .withClientOptions(new ClientConfig().withNodes(Env.NODE))
+                .withSecretManager(
+                        new StrongholdSecretManager(Env.STRONGHOLD_PASSWORD, null, Env.STRONGHOLD_SNAPSHOT_PATH))
+                .withCoinType(CoinType.Shimmer)
+        );
+
+        // Get account and sync it with the registered node to ensure that its balances are up-to-date.
+        AccountHandle a = wallet.getAccount(new AccountAlias(Env.ACCOUNT_NAME));
+        AccountBalance balance = a.syncAccount(new SyncAccount().withOptions(new SyncOptions()));
```
kwek20, line 27: So we sync here twice, and in the requestFundsFromFaucet as well, is all that necesary?

samuel-rufi: only `a.syncAccount()` is synching here, so its not twice
and the inner of requestFundsFromFaucet() needs to sync too, to check if funding worked

bindings/java/lib/src/main/java/org/iota/types/RequiredStorageDeposit.java
samuel-rufi: good point, fixed!

bindings/java/examples/src/RequestFundsFromFaucet.java
samuel-rufi: the sync after is not really needed true (since the inner of requestFundsFromFaucet() is already doing it)
but better illustrate it to the user

kwek20: dont we auto sync with a delay on creating account? :) but yea to illustrate the point is good enough for me

kwek20 (dismissed):

thibault-martinez (dismissed):

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
```diff
@@ -556,6 +557,29 @@ public Transaction createAliasOutput(CreateAliasOutput options) throws WalletExc
...
+                    e.printStackTrace();
+                }
+            } else
+                return;
+        }
+
+        if(syncAccount(new SyncAccount()).getBaseCoin().getAvailable() < 1000000)
+            throw new NoFundsReceivedFromFaucetException();
+    }
+
```
Thoralf-M: Why not just send the faucet request and return the response?
Or if you want to add some account sync, you also need to accept the sync options, otherwise it can happen that you sync outputs which usually don't get synced (for example exchange only wants to sync basic outputs with no further unlock conditions, with the default they would now also get other outputs when calling this command)

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
```diff
@@ -556,6 +557,29 @@ public Transaction createAliasOutput(CreateAliasOutput options) throws WalletExc
...
 
+    /**
+     * Asks the faucet for funds.
+     *
+     * @throws NoFundsReceivedFromFaucetException when the faucet didn't fund the address.
+     */
+    public void requestFundsFromFaucet(RequestFundsFromFaucet options) throws WalletException, NoFundsReceivedFromFaucetException {
+        int maxAttempts = 5;
+        for(int i = 0; i < maxAttempts; i++) {
+            if(syncAccount(new SyncAccount()).getBaseCoin().getAvailable() < 1000000) {
```
Thoralf-M: Who says that the user doesn't want to get more coins even if he has more than 1000000? :trollface:

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
samuel-rufi: When requesting there is no guarantee that the faucet will process the request soon.
And the Faucet also doesn't answer with a nice message that we could return to the user.
All this is really bad for the UX. Calling only the function and returning the current HTML is bad IMO. Imo we need some improved faucet REST API.
Thats the reason why I added syncing here with some retries in case there was no funding transaction.  The idea is to make the requestFundsFromFaucet() more reliable.

The user shouldn't use the requestFundsFromFaucet() to sync his account. He shouldn't care and know about the synching in the requestFundsFromFaucet() .
When he wants to sync his account he can do so after.

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
samuel-rufi: The faucet funds in 10 SMR right? Maybe better if we do like:

`if(syncAccount(new SyncAccount()).getBaseCoin().getAvailable() < 10000000)`

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
Thoralf-M: But you're calling the sync method internally and then can cause issues

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
Thoralf-M: My point is, let's say you have 1000000 already and you call this function, why should it return without requesting more funds from the faucet?
If you want to check with the amount it should be more like `int balanceBeforeFaucetCall = account.getBaseCoin().getAvailable()` and then later just check if the balance increased

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
samuel-rufi: why can it cause issues?

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
samuel-rufi: okay but the faucet won't fund in case there are already enough funds on the address.

> int balanceBeforeFaucetCall = account.getBaseCoin().getAvailable() and then later just check if the balance 

Won't return a change then, and after 5 try attempts (since of no change) it will throw an error.
It shouldn't throw an error in this case.
But to know that we need to check if there are already enough funds?

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
Thoralf-M: > it can happen that you sync outputs which usually don't get synced (for example exchange only wants to sync basic outputs with no further unlock conditions, with the default they would now also get other outputs when calling this command)

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
Thoralf-M: Then just return no error if you have balance
> The faucet funds in 10 SMR right?

I think that's what you get by default after the initial 1000, but if you run your own faucet, you can configure this value to something different, so it shouldn't be hardcoded I think

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
samuel-rufi: I'm now passing the synch options parameter, do you think this is fine like this?

Thoralf-M (approved):

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
Thoralf-M: Yes, this should be fine

bindings/java/lib/src/main/java/org/iota/types/AccountHandle.java
samuel-rufi: > Then just return no error if you have balance
>  but if you run your own faucet, you can configure this value to something different, so it shouldn't be hardcoded I think

so to check for enough balance we need to do a check like `balance > targetAddressBalance`
do you think its fine with latest changes?

thibault-martinez (approved):

kwek20 (approved):

Issue #1644, opened 2022-12-02
Add Swift bindings to CI


thibault-martinez:

thibault-martinez:
Closed by https://github.com/iotaledger/wallet.rs/pull/1667

PR #1667, opened 2022-12-12
Added Swift build on gh actions

kwek20:
# Description of change

Adds building of the swift xcode project for swift bindings to GH actions.

q: Do we always want to run GH actions on `ci/` branches or shall i remove that again?

fixes issue #1644

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -6,6 +6,7 @@ on:
       - main
       - dev
       - develop
+      - ci/**
   pull_request:
     branches:
       - dev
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -194,6 +195,79 @@ jobs:
...
+          toolchain: stable
+          override: true
+
+      # Temporary fix for "ValueError: invalid mode: 'rU' while trying to load binding.gyp"
+      # See <PR link>
+      # This can be removed when "prebuild" updates "node-gyp"
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
rajivshah3: This is just for the Node.js bindings (because of `node-gyp`) so I think it can be removed for Swift

.github/workflows/build-and-test.yml
```diff
@@ -194,6 +195,79 @@ jobs:
...
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --all-features --release --manifest-path bindings/swift/Cargo.toml
+
+      - name: Move build library to xcode folder
+        run: mv "$GITHUB_WORKSPACE/bindings/swift/target/release/libiota_wallet.dylib" "$GITHUB_WORKSPACE/bindings/swift/xcode/IotaWallet/iota_wallet"
+
+      - name: Select Xcode
+        run: sudo xcode-select -switch /Applications/Xcode_13.0.app && /usr/bin/xcodebuild -version
```
rajivshah3: I think this might be better so we don't have to change this with each Xcode release (the latest stable version is 14.1):

```suggestion
      - uses: maxim-lobanov/setup-xcode@v1
        with:
          xcode-version: latest-stable
```

.github/workflows/build-and-test.yml
```diff
@@ -194,6 +195,79 @@ jobs:
...
+      
+      - name: Build project
+        run: xcodebuild build -scheme IotaWallet -project IotaWallet.xcodeproj -destination 'platform=macOS' | xcpretty && exit ${PIPESTATUS[0]}
```
rajivshah3: I think this might be a bit better and is more common:

```suggestion
        run: |
          set -o pipefail
          xcodebuild build -scheme IotaWallet -project IotaWallet.xcodeproj -destination 'platform=macOS' | xcpretty
```

rajivshah3: It turns out that `$PIPESTATUS` is actually just available on `bash`, not `sh`, so it should either have `set -o pipefail` added before or we should add `shell: bash` ([which adds `set -eo pipefail`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference))

Thoralf-M (approved):

rajivshah3 (approved):

PR #1681, opened 2022-12-19
feat: query all node's events

### LINKED ISSUE #1670: [Task]: Add API call that gets the events on the client nodes
## Description

Add an API call that fetches the events from the nodes that are registered in the client options.

## Motivation

For the governance feature we would like to show the events that are available on the given node.

## Requirements

Write a list of what you want this feature to do.

1. Add `getUnregisteredEventsFromClient(filterOptions?: FilterOptions)` API call
2. Add filter options to get only the voting events
3. Returns an array of `Event`

No.

maxwellmattryan:
# Description of change

This is needed for Firefly to automatically populate the Governance dashboard with proposals from the client option's node(s).

Closes #1670 

## Type of change

Modified the participation example to query a local node with both voting and staking events.

## Change checklist


thibault-martinez (changes requested):
Please bump the client revision

src/account_manager/operations/participation.rs
```diff
@@ -62,6 +62,19 @@ impl AccountManager {
...
+    pub async fn get_participation_events_from_client(
+        &self,
+        event_type: Option<ParticipationEventType>,
+    ) -> crate::Result<Vec<EventId>> {
+        let accounts = self.accounts.read().await;
+        Ok(if let Some(account) = accounts.first() {
+            let events = account.client.events(event_type).await?;
+            events.event_ids
+        } else {
+            vec![]
```
Thoralf-M: If there is no client we should build it, similar to here
https://github.com/iotaledger/wallet.rs/blob/6b3cdfed7bf4960b2c682a4b10353df58711f294/src/account_manager/mod.rs#L165-L169

src/account_manager/operations/participation.rs
```diff
@@ -62,6 +62,19 @@ impl AccountManager {
...
+        &self,
+        event_type: Option<ParticipationEventType>,
+    ) -> crate::Result<Vec<EventId>> {
+        let accounts = self.accounts.read().await;
+        let client = match accounts.first() {
+            Some(account) => account.client.clone(),
+            None => self.client_options.read().await.clone().finish()?,
+        };
+        let events = client.events(event_type).await?;
+        Ok(events.event_ids)
```
thibault-martinez, line 86:
```suggestion

        Ok(events.event_ids)
```

src/account_manager/operations/participation.rs
```diff
@@ -62,6 +62,20 @@ impl AccountManager {
             .collect())
     }
 
+    pub async fn get_participation_events_from_client(
```
thibault-martinez: Please add a small documentation for the methods since it's a public one.
Also, maybe the name is confusing since we are not returning events but events ids.
I'd drop the `_client` and call it `get_participation_event(s)_ids` ?

src/account_manager/operations/participation.rs
```diff
@@ -62,6 +62,20 @@ impl AccountManager {
...
+        &self,
+        event_type: Option<ParticipationEventType>,
+    ) -> crate::Result<Vec<EventId>> {
+        let accounts = self.accounts.read().await;
+        let client = match accounts.first() {
+            Some(account) => account.client.clone(),
```
thibault-martinez: Cloning a client has some consequences (spawning tasks etc, ...) so the less we do it the better. Here, there is an opportunity to avoid cloning it by directly calling `events` in both branches. It is a bit of a duplicate but it'll be more efficient.

src/account_manager/operations/participation.rs
maxwellmattryan: I'll modify the name, but it might be worth adding a second argument to the function `nodes: Option<Vec<Node>>` then if no nodes are provided we go with the client options one. This makes the function a bit more usable. If we don't have this second argument then IMO it's slightly clearer to leave the `_from_client` in the name.

src/account_manager/operations/participation.rs
thibault-martinez: Small mistake, it doesn't actually spawn but it's still a clone so I'd rather do this anyway

maxwellmattryan: I was only able to remove the `clone()` call for the first branch :thinking:

thibault-martinez: Yeah that's the one I was speaking about, you can't remove the second one

src/account_manager/operations/participation.rs
thibault-martinez: Even without an additional parameter I feel like it's redundant. It's quite obvious it calls the client, if we follow the logic we would need to add that prefix to a LOT of wallet methods...

maxwellmattryan: That's fair. In that case, I'll add the second optional argument.

maxwellmattryan: That's fair. I'll change it and not add the optional parameter, since it should be no problem to add later should we decide we need it :+1:

thibault-martinez:
> Please add a .change. You also need to bump the client revision I guess.

You still need to address this :)

thibault-martinez:
Also please add this in the changelog 🙏🏻

src/account_manager/operations/participation.rs
Tuditi: Why is the function returning `event_ids` and not `events`?

kwek20 (dismissed):

src/account_manager/operations/participation.rs
maxwellmattryan: There are a few reasons:
1. The node returns just the events' IDs, not the corresponding event information / payload.
2. The events will have to be "registered" anyway, which requires an ID and a node. This function, `register_participation_events` returns `Result<Event>` anyway, so Firefly will have the data eventually.

Alternatively, we could expose an additional method e.g. `register_participation_events_of_node` that would return `Result<Vec<Event>>`, but since this would only be particularly useful for the official events on official nodes, it is alright IMO if Firefly just handles these extra steps.

Thoralf-M (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

kwek20 (approved):

PR #1702, opened 2023-01-05
Use recommended approach for caching CARGO_HOME

rajivshah3:
# Description of change

- Use [recommended approach](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci) for caching `CARGO_HOME`
  - Files from `CARGO_HOME` are now in a single cache artifact instead of two
  - By removing duplicates and things that Cargo can generate quickly, the cache artifact size is reduced by ~50 MB
- Fix skipped steps in Java workflow on Windows
- Remove trigger for branches beginning with "ci"
  - This causes builds to be created twice (once for the PR, once for the push)
- Use stable toolchain to compile `udeps` so that we can cache it

N/A

## Type of change

Workflows pass

## Change checklist


.github/workflows/build-and-test.yml
```diff
@@ -153,27 +146,19 @@ jobs:
...
         with:
-          path: ~/.cargo/git
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
           # Add date to the cache to keep it up to date
-          key: ${{ matrix.os }}-stable-cargo-index-${{ hashFiles('**/Cargo.lock') }}-${{ env.CURRENT_DATE }}
+          key: ${{ matrix.os }}-stable-cargo-home-${{ env.CURRENT_DATE }}
```
Thoralf-M, line 158: Do you think it's useful to have this date here? We had some doubts, but I'm not 100% sure what effect it has

rajivshah3: I think I was wrong in adding the `hashFiles('**/Cargo.lock')` because we don't use a lockfile for the crate. Since GitHub Actions only saves a new cache when there isn't a hit for the exact key ([see docs](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache)), we need a key that changes. Usually the `Cargo.lock` would be good but we don't have one for the crate. If we keep using something like `${{ matrix.os }}-stable-cargo-home`, the cache will never be updated as there will always be a hit on the exact key. As an alternative though, we could also use something like week number (e.g. `2022-01` for Jan 1-7) if creating a new cache every day is excessive. The main issue is that not having a `Cargo.lock` for the crate means that caching `CARGO_HOME` becomes a bit more difficult

kwek20: I would say that current date is fine, as thats still a day. Usually we run a pr a few times per day, and then its merged or takes a while. so 1 week wouldnt really improve much

.github/workflows/udeps.yml
```diff
@@ -28,17 +28,19 @@ jobs:
...
-      uses: actions/cache@v3
-      with:
-        path: ~/.cargo/git
-        key: ubuntu-latest-nightly-cargo-index-${{ hashFiles('**/Cargo.toml') }}
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+        # Add date to the cache to keep it up to date
```
thibault-martinez, line 42: What's this comment about ? I don't see any date ?

rajivshah3: Ooof good catch. The way we're doing this actually isn't the best, because we're compiling `udeps` with the nightly toolchain (so caching might be useless because the toolchain may change every night and cause `udeps` to be rebuilt). It only needs to be _run_ with the nightly toolchain but it can be built with stable. I'll re-work this a bit

Thoralf-M (commented):
What about `java_binding_publish` and `private_tangle_tests`? Should they also be updated?

thibault-martinez:
Hi @rajivshah3 , can you reopen this PR on the SDK repository please ?

PR #1715, opened 2023-01-09
Add and use TransactionOptionsDto

### LINKED ISSUE #1692: TransactionOptions should use TaggedDataPayloadDto in message interface
## Bug description

Currently TaggedDataPayload is required, which would look like `"taggedDataPayload":{"tag":{"inner":[1],"bounded":null},"data":{"inner":[1],"bounded":null}}` (tag and data both being [1])

## Expected behaviour

The dto should be used in the message interface (a TransactionOptionsDto should be added I guess), so the bindings can provide it in this format
`"taggedDataPayload": {
            "type": 5,
            "tag": "0x01",
            "data": "0x01"
          }`

Thoralf-M:
# Description of change

Add and use TransactionOptionsDto so the types are consistent

Fixes #1692 

## Type of change

## Change checklist

### DIFF
--- a/src/message_interface/message_handler.rs
+++ b/src/message_interface/message_handler.rs
@@ -36,6 +36,7 @@ use crate::{
         operations::transaction::{
             high_level::{create_alias::AliasOutputOptions, minting::mint_native_token::MintTokenTransactionDto},
             prepare_output::OutputOptions,
+            TransactionOptions,
         },
         types::{AccountBalanceDto, AccountIdentifier, TransactionDto},
         OutputDataDto,
@@ -216,7 +217,7 @@ impl WalletMessageHandler {
             }),
             Message::SetClientOptions(options) => {
                 convert_async_panics(|| async {
-                    self.account_manager.set_client_options(*options.clone()).await?;
+                    self.account_manager.set_client_options(*options).await?;
                     Ok(Response::Ok(()))
                 })
                 .await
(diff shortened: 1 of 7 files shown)


src/message_interface/message_handler.rs
```diff
@@ -535,7 +539,17 @@ impl WalletMessageHandler {
             }
             AccountMethod::BurnNft { nft_id, options } => {
                 convert_async_panics(|| async {
-                    let transaction = account_handle.burn_nft(NftId::try_from(&nft_id)?, options).await?;
+                    let transaction = account_handle
+                        .burn_nft(
+                            NftId::try_from(&nft_id)?,
+                            match options {
```
thibault-martinez: For this and all of them, looks like you could just map instead of match ?

Thoralf-M: But with the map I can't do the error handling? I think there was some way to do the same with less code, but couldn't find it

thibault-martinez: Ah yeah right I missed the `?`. Maybe just `transpose()?` after the `map` ?

Thoralf-M: Yes that was what I searched, thanks!

thibault-martinez (dismissed):

thibault-martinez (approved):

kwek20 (approved):
Today i learned about `transpose()` 🤯

PR #1718, opened 2023-01-09
Wallet tester workflow improvements

rajivshah3:
# Description of change
- Remove unused `yq` action
- Use `actions/checkout` to clone Hornet repo
- Run `cleanup.sh` as part of private tangle tear-down
- Cancel wallet tester after 15 min
- Always tear down the private tangle even if previous steps fail
- Avoid printing image pull progress as it fills up the logs

N/A

## Type of change

- Refactor

Workflow succeeds

## Change checklist

### DIFF
--- a/.github/actions/private-tangle/setup/action.yml
+++ b/.github/actions/private-tangle/setup/action.yml
@@ -32,7 +32,7 @@ runs:
         # Start Tangle
         sudo ./cleanup.sh
         sudo ./bootstrap.sh
-        sudo ./run.sh -d
+        sudo ./run.sh -d --quiet-pull
       working-directory: hornet/private_tangle
 
     - name: Wait for tangle to start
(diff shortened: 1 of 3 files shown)


.github/actions/private-tangle/setup/action.yml
```diff
@@ -3,7 +3,12 @@ description: 'Setup a private tangle'
 runs:
   using: "composite"
   steps:
-    - uses: mikefarah/yq@master
```
rajivshah3, line 6: This action is intended for running commands with `yq` but no inputs were supplied, so it wasn't doing anything. On the GitHub-hosted runners, [`yq` is already installed](https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/yq.sh) so we don't need to worry about installing it ourselves

Thoralf-M (approved):

thibault-martinez (approved):
Approving for the changes to the private tangle setup that can be useful later but I think I'll remove the wallet tester entirely as we won't be using it like this anymore. Thanks anyway!

PR #1728, opened 2023-01-13
Feat: rocksdb as optional storage

kwek20:
I would leave out rocksdb as default completely, but some tests rely on data storage persistence.
An option is to make the Memory hashmap global, but i dont think thats worth the trouble

This is a pre for wasm storage in the browser (LocalStorage)


Cargo.toml
```diff
@@ -39,10 +39,10 @@ fern-logger = { version = "0.5.0", default-features = false }
 
 [features]
 # only default for now, should later be replaced by stronghold since stronghold is more secure
-default = [ "mnemonic", "storage", "stronghold" ]
+default = [ "mnemonic", "storage", "rocksdb", "stronghold" ]
 ledger_nano = [ "iota-client/ledger_nano" ]
 stronghold = [ "iota-client/stronghold" ]
-storage = [ "rocksdb" ]
+storage = []
```
Thoralf-M, line 45: Why not keep rocksdb here, if we have to keep it (at least for now)?

thibault-martinez: If we keep it, we couldn't be able to say that we want storage but not rocksdb when compiling for wasm

src/storage/manager.rs
```diff
@@ -20,8 +20,18 @@ pub(crate) enum ManagerStorage {
     #[cfg(feature = "stronghold")]
     /// Stronghold storage.
     Stronghold,
+    #[cfg(feature = "rocksdb")]
```
thibault-martinez: I know it was like that for stronghold already but can you put the attribute after the doc comment please ? That's what we mostly do.

src/storage/constants.rs
```diff
@@ -4,9 +4,17 @@
 /// The default storage path.
 pub const DEFAULT_STORAGE_PATH: &str = "./storage";
 
+#[cfg(feature = "rocksdb")]
```
thibault-martinez: Same

src/storage/adapter/mod.rs
```diff
@@ -22,3 +27,32 @@ pub trait StorageAdapter: std::fmt::Debug {
     /// Removes a record from the storage.
     async fn remove(&mut self, key: &str) -> crate::Result<()>;
 }
+
+#[derive(Debug, Default)]
```
thibault-martinez: same

src/storage/adapter/mod.rs
```diff
@@ -22,3 +27,32 @@ pub trait StorageAdapter: std::fmt::Debug {
...
+/// A storage adapter that stores data in memory.
+pub struct Memory(HashMap<String, String>);
+
+#[async_trait::async_trait]
+impl StorageAdapter for Memory {
```
thibault-martinez: Can you put memory in a dedicated module please ?

Cargo.toml
Thoralf-M: But the way it's in this PR, we also can't use it without default features but with the storage feature, because rocksdb isn't added as new feature?

kwek20: any optional dependency automatically is a feature, so you can just use it like normal

Thoralf-M: Ah nice, I had forgotten that

src/storage/adapter/mod.rs
kwek20: I left it in mod as its sort of a testing/fallback adapter, but yea i will move it

thibault-martinez (changes requested):
Clippy

Thoralf-M (dismissed):

src/storage/manager.rs
```diff
@@ -17,11 +17,21 @@ use crate::{
...
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub(crate) enum ManagerStorage {
-    #[cfg(feature = "stronghold")]
     /// Stronghold storage.
+    #[cfg(feature = "stronghold")]
     Stronghold,
     /// RocksDB storage.
+    #[cfg(feature = "rocksdb")]
     Rocksdb,
+    /// Storage backed by a Map in memory, for testing
```
thibault-martinez:
```suggestion
    /// Storage backed by a Map in memory.
```
Removing because it may not only be for testing, browser may use that in the future.

src/storage/adapter/mod.rs
```diff
@@ -2,10 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// RocksDB storage adapter.
+#[cfg(feature = "rocksdb")]
 pub mod rocksdb;
```
thibault-martinez, line 8:
```suggestion
```

src/storage/adapter/memory.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2021 IOTA Stiftung
```
thibault-martinez:
```suggestion
// Copyright 2023 IOTA Stiftung
```

src/storage/adapter/mod.rs
```diff
@@ -2,10 +2,16 @@
...
+pub mod memory;
```
thibault-martinez:
```suggestion
pub mod memory;
pub mod rocksdb;
```

src/storage/adapter/mod.rs
```diff
@@ -2,10 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// RocksDB storage adapter.
+#[cfg(feature = "rocksdb")]
+pub mod memory;
 pub mod rocksdb;
```
thibault-martinez:
```suggestion
pub mod memory;
#[cfg(feature = "rocksdb")]
pub mod rocksdb;
```

src/storage/adapter/mod.rs
```diff
@@ -2,10 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// RocksDB storage adapter.
+pub mod memory;
+#[cfg(feature = "rocksdb")]
```
thibault-martinez:
```suggestion
pub mod memory;
/// RocksDB storage adapter.
#[cfg(feature = "rocksdb")]
```

src/storage/manager.rs
```diff
@@ -17,11 +17,21 @@ use crate::{
...
     Rocksdb,
+    /// Storage backed by a Map in memory.
+    Memory,
+}
+
+pub(crate) const fn default_storage() -> ManagerStorage {
+    #[cfg(feature = "rocksdb")]
+    return ManagerStorage::Rocksdb;
+    #[cfg(not(feature = "rocksdb"))]
+    ManagerStorage::Memory
```
thibault-martinez: Can this not be an impl Default ?

kwek20: Not for an enum no, unfortunately 😢

thibault-martinez: What do you mean ? What's the issue with 
```rust
impl Default for ManagerStorage {
    fn default() -> ManagerStorage {
        #[cfg(feature = "rocksdb")]
        return ManagerStorage::Rocksdb;
        #[cfg(not(feature = "rocksdb"))]
        ManagerStorage::Memory
    }
}
```

kwek20: oh.. why did that not come up in my brain... i was thinking about derive default and cant do that in enums haha

thibault-martinez: You actually can https://rust-lang.github.io/rfcs/3107-derive-default-enum.html, I'm just not sure of the syntax together with feature gates but it should be doable. But fine to impl anyway.

Thoralf-M (dismissed):

src/storage/adapter/memory.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Debug, Default)]
+pub struct Memory(HashMap<String, String>);
+
+#[async_trait::async_trait]
+impl StorageAdapter for Memory {
+    fn id(&self) -> &'static str {
+        STORAGE_ID
+    }
+
+    async fn get(&self, key: &str) -> crate::Result<String> {
```
thibault-martinez, line 22: This one is missing doc

thibault-martinez, line 22:
```suggestion
    /// Gets a record from the storage.
    async fn get(&self, key: &str) -> crate::Result<String> {
```

src/storage/adapter/memory.rs
```diff
@@ -0,0 +1,42 @@
...
+        self.0.get(key).ok_or_else(|| storage_err(key)).cloned()
+    }
+
+    /// Saves or updates a record on the storage.
+    async fn set(&mut self, key: &str, record: String) -> crate::Result<()> {
+        self.0.insert(key.to_string(), record);
+        Ok(())
+    }
+
+    /// Batch write.
```
thibault-martinez:
```suggestion
    /// Batch writes records to the storage.
```

thibault-martinez (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

PR #1739, opened 2023-01-18
Update addresses eventually when client config changes

### LINKED ISSUE #1701: smr1 addres generated in testnet
## Bug description

smr1 address generated on testnet node  ("https://api.testnet.shimmer.network")
## Rust version

Which version of Rust are you running?

- Rust version: java bindings

## Version

Which version of the library are you using?

- Version number, commit, or branch: 

## Hardware specification

What hardware are you using?

- Operating system: windows
- RAM: 16 
- Cores: 8
- Device: normal pc

## Steps To reproduce the bug

```
public static void main(String[] args) throws WalletException {
		// Build the wallet.
		Wallet wallet;
		try {
			wallet = new Wallet(new WalletConfig()
					.withClientOptions(new ClientConfig().withNodes("https://api.testnet.shimmer.network"))
					.withSecretManager(new StrongholdSecretManager("PASSWORD_FOR_ENCRYPTION", null, "example-wallet"))
					.withCoinType(CoinType.Shimmer));
			wallet.storeMnemonic(DEFAULT_DEVELOPMENT_MNEMONIC);

			// Create an account.
			AccountHandle a = wallet.createAccount("Alice4");

			// Print the account.
			System.out.println(a);

			
		} catch (InitializeWalletException e) {
			// TODO Auto-generated catch block
			e.printStackTrace();
		}

	}
```

## Expected behaviour

create testnet address 

## Actual behaviour
creates mainnet address

## Errors

Paste any errors that you see, including logs, errors, or screenshots.

samuel-rufi:
Fixes https://github.com/iotaledger/wallet.rs/issues/1701


src/account_manager/builder.rs
```diff
@@ -154,53 +156,63 @@ impl AccountManagerBuilder {
...
             return Ok(AccountManager {
                 #[cfg(not(feature = "events"))]
-                accounts: Arc::new(RwLock::new(
-                    accounts
-                        .into_iter()
-                        .map(|a| AccountHandle::new(a, client.clone(), secret_manager.clone(), storage_manager.clone()))
-                        .collect(),
-                )),
+                accounts: Arc::new(RwLock::new(account_handles)),
                 #[cfg(feature = "events")]
```
Thoralf-M: `#[cfg(feature = "events")]` and `#[cfg(not(feature = "events"))]` can be removed here, no?

src/account_manager/builder.rs
```diff
@@ -154,53 +156,63 @@ impl AccountManagerBuilder {
...
+                        .ok_or(crate::Error::MissingParameter("secret_manager"))?,
+                    self.coin_type
+                        .ok_or(crate::Error::MissingParameter("coin_type (IOTA: 4218, Shimmer: 4219)"))?,
+                ),
             };
 
-            let client = client_options.clone().finish()?;
+            self.client_options.replace(client_options.clone());
+            self.secret_manager.replace(secret_manager.clone());
+            self.coin_type.replace(coin_type);
```
Thoralf-M: In the None case above these values have to be defined already in self, so actually we only need to set it in the `Some(ref data) => {` case

src/account_manager/builder.rs
samuel-rufi: line 215 we pass the event_emitter, so we need the `#[cfg(feature = "events")]` in this case, don't we?

samuel-rufi: good catch, fixed it

src/account_manager/builder.rs
samuel-rufi: fixed

src/account_manager/builder.rs
```diff
@@ -134,18 +134,20 @@ impl AccountManagerBuilder {
...
                             .client_options
+                            .clone()
                             .ok_or(crate::Error::MissingParameter("client_options"))?,
                     };
                     let secret_manager = match self.secret_manager {
-                        Some(secret_manager) => secret_manager,
+                        Some(ref secret_manager) => secret_manager.clone(),
                         None => data
                             .secret_manager
+                            .clone()
```
Thoralf-M: Do we actually need the refs and clones here now since the replace got moved?

samuel-rufi: yes couldn't get it working when removing that

bindings/java/lib/native/src/lib.rs
```diff
@@ -73,7 +73,8 @@ macro_rules! env_assert {
 pub extern "system" fn Java_org_iota_api_NativeApi_initLogger(env: JNIEnv, _class: JClass, command: JString) {
     // This is a safety check to make sure that the JNIEnv is not in an exception state.
     env_assert!(env,());
-    jni_err_assert!(env, init_logger(string_from_jni!(env, command, ())), ());
+    let ret = init_logger(string_from_jni!(env, command, ()));
```
samuel-rufi, line 76:
```suggestion
    // Must not be a one-liner, otherwise the macro will fail.
    let ret = init_logger(string_from_jni!(env, command, ()));
```

src/account_manager/builder.rs
Thoralf-M: @samuel-rufi didn't want to annoy you, pls review https://github.com/samuel-rufi/wallet.rs/pull/123 :trollface:

samuel-rufi: Lgtm

src/account_manager/builder.rs
```diff
@@ -130,93 +129,93 @@ impl AccountManagerBuilder {
...
+            }
         }
 
         Ok(AccountManager {
-            accounts: Arc::new(RwLock::new(Vec::new())),
+            accounts: Arc::new(RwLock::new(account_handles)),
             background_syncing_status: Arc::new(AtomicUsize::new(0)),
             client_options: Arc::new(RwLock::new(
                 self.client_options
+                    .clone()
```
Thoralf-M:
```suggestion
```

src/account_manager/builder.rs
```diff
@@ -130,90 +129,89 @@ impl AccountManagerBuilder {
...
-                event_emitter,
-                storage_options,
-                storage_manager,
-            });
+        if self.coin_type.is_none() {
+            let coin_type = read_manager_builder
+                .and_then(|data| data.coin_type)
+                .ok_or(crate::Error::MissingParameter("coin_type (IOTA: 4218, Shimmer: 4219)"))?;
+
+            // Update self so it get sused and stored again
```
Thoralf-M:
```suggestion
            // Update self so it gets used and stored again
```

src/account_manager/builder.rs
```diff
@@ -130,90 +129,89 @@ impl AccountManagerBuilder {
...
+        let read_manager_builder: Option<AccountManagerBuilder> = None;
+
+        // prioritise provided client_options and secret_manager over stored ones
+        let new_provided_client_options = if self.client_options.is_none() {
+            let loaded_client_options = read_manager_builder
+                .as_ref()
+                .and_then(|data| data.client_options.clone())
+                .ok_or(crate::Error::MissingParameter("client_options"))?;
+
+            // Update self so it get sused and stored again
```
Thoralf-M:
```suggestion
            // Update self so it gets used and stored again
```

src/account_manager/builder.rs
```diff
@@ -130,90 +129,89 @@ impl AccountManagerBuilder {
...
+            true
+        };
+
+        if self.secret_manager.is_none() {
+            let secret_manager = read_manager_builder
+                .as_ref()
+                .and_then(|data| data.secret_manager.clone())
+                .ok_or(crate::Error::MissingParameter("secret_manager"))?;
+
+            // Update self so it get sused and stored again
```
Thoralf-M:
```suggestion
            // Update self so it gets used and stored again
```

src/account_manager/builder.rs
```diff
@@ -130,90 +129,89 @@ impl AccountManagerBuilder {
...
+                    #[cfg(feature = "events")]
+                    event_emitter.clone(),
+                    #[cfg(feature = "storage")]
+                    storage_manager.clone(),
+                )
+            })
+            .collect::<_>();
+
+        // If the manager builder is not set, it means the user provided it and we need to update the addresses.
+        // In the other case it was loaded from the database and addresses are up to date.
```
Thoralf-M, line 215:
```suggestion
        // If they are new, we need to update the accounts because the bech32 HRP might have changed.
        // In the other case it was loaded from the database and addresses are up to date.
```

Thoralf-M (dismissed):

src/account_manager/builder.rs
```diff
@@ -121,102 +119,103 @@ impl AccountManagerBuilder {
...
+            }
         }
 
         Ok(AccountManager {
-            accounts: Arc::new(RwLock::new(Vec::new())),
+            accounts: Arc::new(RwLock::new(account_handles)),
             background_syncing_status: Arc::new(AtomicUsize::new(0)),
             client_options: Arc::new(RwLock::new(
                 self.client_options
+                    .clone()
```
Thoralf-M:
```suggestion
```

Thoralf-M (changes requested):
Currently it can still fail because the accounts aren't stored in `update_account_with_new_client()`, we should add some code there which stores the accounts to db at the end of the function like here https://github.com/iotaledger/wallet.rs/blob/f118f02871505488984b17df892f8ab8f47dc5e2/src/account/update.rs#L171-L175

Example I used for testing:

```Rust
use std::{env, path::PathBuf};

use dotenv::dotenv;
use iota_wallet::{
    account_manager::AccountManager,
    iota_client::constants::SHIMMER_COIN_TYPE,
    secret::{stronghold::StrongholdSecretManager, SecretManager},
    ClientOptions, Result,
};

#[tokio::main]
async fn main() -> Result<()> {
    dotenv().ok();

    let mut secret_manager = StrongholdSecretManager::builder()
        .password(&env::var("STRONGHOLD_PASSWORD").unwrap())
        .build(PathBuf::from("wallet.stronghold"))?;

    let mnemonic = env::var("NON_SECURE_USE_OF_DEVELOPMENT_MNEMONIC").unwrap();
    let _ = secret_manager.store_mnemonic(mnemonic).await;

// manually switch between https://api.testnet.shimmer.network and https://api.shimmer.network
    let client_options = ClientOptions::new().with_node(&"https://api.testnet.shimmer.network")?;

    let manager = AccountManager::builder()
        .with_secret_manager(SecretManager::Stronghold(secret_manager))
        // comment out to see what's happening when loaded from db only
        .with_client_options(client_options)
        .with_coin_type(SHIMMER_COIN_TYPE)
        .finish()
        .await?;

    let account = manager.create_account().finish().await?;

    println!("New: {}", &account.addresses().await?[0].address().to_bech32());

    let accounts = manager.get_accounts().await?;
    for account in accounts {
        println!("{}", &account.addresses().await?[0].address().to_bech32());
    }

    Ok(())
}
```

src/account/update.rs
```diff
@@ -274,6 +274,13 @@ impl AccountHandle {
         for address in &mut account.internal_addresses {
             address.address.bech32_hrp = bech32_hrp.clone();
         }
+
+        #[cfg(feature = "storage")]
+        {
+            log::debug!("[SYNC] storing account {} with new synced data", account.alias());
```
Thoralf-M: We didn't sync there, so maybe something like
```suggestion
            log::debug!("[SYNC] storing account {} after updating it with new client options", account.alias());
```

samuel-rufi: should be fine like this now

Thoralf-M (dismissed):

src/account_manager/builder.rs
```diff
@@ -134,90 +134,89 @@ impl AccountManagerBuilder {
...
-                        self.coin_type
-                            .ok_or(crate::Error::MissingParameter("coin_type (IOTA: 4218, Shimmer: 4219)"))?,
-                    )
-                }
-            };
-
-            let client = client_options.clone().finish()?;
-
-            let accounts = storage_manager.lock().await.get_accounts().await.unwrap_or_default();
+        let read_manager_builder = storage_manager.lock().await.get_account_manager_data().await.ok();
```
thibault-martinez: Not a big fan of this `.ok()`, it's silencing errors

src/account_manager/builder.rs
```diff
@@ -134,90 +134,89 @@ impl AccountManagerBuilder {
...
+            let coin_type = read_manager_builder
+                .and_then(|data| data.coin_type)
+                .ok_or(crate::Error::MissingParameter("coin_type (IOTA: 4218, Shimmer: 4219)"))?;
+
+            // Update self so it gets used and stored again
+            self.coin_type.replace(coin_type);
+        }
+
+        #[cfg(feature = "storage")]
+        // Store account manager data in storage
```
thibault-martinez, line 180: Can you switch comment and attribute pls

src/account_manager/builder.rs
```diff
@@ -136,7 +136,7 @@ impl AccountManagerBuilder {
         .await?;
 
         #[cfg(feature = "storage")]
-        let read_manager_builder = storage_manager.lock().await.get_account_manager_data().await.ok();
+        let read_manager_builder = Some(storage_manager.lock().await.get_account_manager_data().await?);
```
Thoralf-M: I think you need to match on the Result because the first time when there is no db you also get an error that the data wasn't found or something and then it shouldn't return the error

samuel-rufi: hmm, when matching the result, what should I do with the Err(x) if not returning it as done in this line?

thibault-martinez: One of the error variant is about the entry not being found I think, so you need to convert that to None, other errors we need to actually rezturn

samuel-rufi: Ah, so basically something like this:

```
        let read_manager_builder = match storage_manager.lock().await.get_account_manager_data().await {
            Ok(x) => Some(x),
            Err(e) => {
                if e.to_string().eq("Key not found") {
                    None
                } else {
                    return Err(e.into());
                }
            }
        };
```

bit unfortunate that the error variant isn't an enum

thibault-martinez: Pretty sure it's `Error::RecordNotFound` but @Thoralf-M can confirm

samuel-rufi: something like this:

```
        #[cfg(feature = "storage")]
        let read_manager_builder = match storage_manager.lock().await.get_account_manager_data().await {
            Ok(x) => Some(x),
            Err(crate::Error::IoError(_)) => None,
            Err(e) => return Err(e),
        };
```

just not sure if the IoError is the correct one atm

thibault-martinez: It's probably not the correct one

Thoralf-M: Yes needs to be `Error::RecordNotFound` :heavy_check_mark:

samuel-rufi: ah im blind

samuel-rufi: there is the     /// Record not found  RecordNotFound(String) in Error struct

samuel-rufi: So for RecordNotFound we need to return None, but I think there is another case:
`{"type":"client","error":"no mnemonic has been stored into the Stronghold vault"}`

searching the variant for that now...

samuel-rufi: works now

thibault-martinez (approved):

Thoralf-M (approved):

PR #1780, opened 2023-01-30
Automate cross-compilation and release of JNI libs for Android

### LINKED ISSUE #1608: Automate cross-compilation and release of JNI libs for Android

samuel-rufi:
Fixes https://github.com/iotaledger/wallet.rs/issues/1608


.github/workflows/java_binding_publish.yml
```diff
@@ -128,4 +136,117 @@ jobs:
           ./gradlew publishToSonatype
           if [ "$RUNNER_OS" == "macOS" ];
           then ./gradlew publishToSonatype -PbuildTarget=aarch64-apple-darwin
-          fi
\ No newline at end of file
+          fi
+
+  release-android:
+    name: Create Android releases
+    needs: release-desktop
```
Thoralf-M: Is this just for `needs.release-desktop.outputs.tag`?

samuel-rufi: yes

Thoralf-M: The version is the one defined in build.gradle? Would be nice to get it directly from there, then you don't need to extract it from the jar file name and you also don't need to wait for the first build to finish before you start the android one

samuel-rufi: good idea, will try something

.github/workflows/java_binding_publish.yml
```diff
@@ -128,4 +136,117 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest]
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: rocksdb-v18
```
kwek20, line 161: We need this branch on iotaledger repo, or at least it tells me it was "renamed to android" so we should be accurate here

samuel-rufi: fixed

kwek20 (approved):

thibault-martinez (approved):

PR #1897, opened 2023-02-27
Fix publish-cli workflow

Thoralf-M:


.github/workflows/publish-cli.yml
```diff
@@ -1,10 +1,6 @@
 name: Publish cli wallet
 
 on:
-  push:
-    branches:
-      - develop
-      - production
   workflow_dispatch:
```
thibault-martinez: Can you do it one line for consistency pls `on: workflow_dispatch`

Thoralf-M: That's not allowed `Nested mappings are not allowed in compact mappings`

Thoralf-M: Nvm, forgot to remove `:`

thibault-martinez (approved):

PR #1906, opened 2023-03-01
Reduce min sync interval

Thoralf-M:
# Description of change

Reduce min sync interval because it looks like some private tangle tests sometimes fails because of it

## Type of change

I tested how long it takes to sign and submit a transaction locally and it was around 6 ms and the other things also take some additional time, so 5 ms should be fine

## Change checklist


wallet/src/account/constants.rs
```diff
@@ -13,7 +13,7 @@ pub(crate) const PARALLEL_REQUESTS_AMOUNT: usize = 500;
 
 /// ms before an account actually syncs with the network, before it just returns the previous syncing result
 /// this is done to prevent unnecessary simultaneous synchronizations
-pub(crate) const MIN_SYNC_INTERVAL: u128 = 200;
+pub(crate) const MIN_SYNC_INTERVAL: u128 = 5;
```
thibault-martinez, line 16: Should we have a value just for testing or do you think it's also fine?

Thoralf-M: I'm not sure, what this should prevent is that if you call .sync() and then .sync() again while the first one hasn't finished, that it doesn't immediately start syncing after it just finished, but just return directly
This depends on the sync mutex and that shouldn't take that long to be available again I guess
A higher value would also help if you call .sync() and .sync() right after it finished, but at the end if it's called twice (the second time after the first finished), then I think it should be okay to sync again

thibault-martinez (dismissed):

wallet/src/account/constants.rs
kwek20, line 16: 5 Ms is basically nothing. Only literally sync(). Sync() would have an effect. Do we we want to keep this at all then?

Thoralf-M: It can also happen if you call it multiple times in parallel and I think for that it makes sense to keep

kwek20 (dismissed):

thibault-martinez (approved):

kwek20 (approved):

PR #1920, opened 2023-03-03
fix: use lower macOS target

Tuditi:
# Description of change

I would like to try to build on a lower macOS target, so we could potentially white screen issues for users with older mac versions.

https://github.com/iotaledger/firefly/issues/4766

## Type of change

None, tests run in the CI

## Change checklist


.github/workflows/covector-version-or-publish.yml
```diff
@@ -101,8 +101,7 @@ jobs:
         if: matrix.os == 'windows-2019'
 
       - name: Set deployment target (macOS)
-        run: echo "MACOSX_DEPLOYMENT_TARGET=10.14" >> $GITHUB_ENV # 10.12 should be supported, but waiting for new rocksDB to test this
-        if: matrix.os == 'macos-11'
```
Thoralf-M, line 105: Shouldn't this stay?

thibault-martinez (dismissed):

Thoralf-M (dismissed):

.github/workflows/build-and-test.yml
```diff
@@ -296,7 +296,7 @@ jobs:
 
     # Set custom deployment target because of rocksdb
     - name: Set deployment target (macOS)
-      run: echo "MACOSX_DEPLOYMENT_TARGET=10.14" >> $GITHUB_ENV 
+      run: echo "MACOSX_DEPLOYMENT_TARGET=10.12" >> $GITHUB_ENV
```
rajivshah3: I recall getting errors with 10.12, but I'll double check to see why this isn't happening anymore:

```
db/version_set.cc:5900:53: error: 'value' is unavailable: introduced in macOS 10.13
              cfd->user_comparator()->Compare(start.value(),
                                                    ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/optional:950:33: note: 'value' has been explicitly marked unavailable here
    constexpr value_type const& value() const&
                                ^
db/version_set.cc:5908:51: error: 'value' is unavailable: introduced in macOS 10.13
              cfd->user_comparator()->Compare(end.value(),
                                                  ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/optional:950:33: note: 'value' has been explicitly marked unavailable here
    constexpr value_type const& value() const&
```

This was with RocksDB though, not `rust-rocksdb`, so maybe that's the issue

kwek20 (dismissed):

.github/workflows/build-and-test.yml
thibault-martinez: @rajivshah3 any news on this?

rajivshah3: Seems like this is still the case for the Node.js bindings. I'm not sure why it's not happening with the Python bindings on CI. There are two groups of errors when I compile the Node.js bindings locally:

1. `introduced in macOS 10.13`: this seems right. The good thing is that based on [this website](https://www.macworld.com/article/673697/what-version-of-macos-can-my-mac-run.html), every computer that supported 10.12 (Sierra) supports 10.13 (High Sierra) so users should be able to update their OS
2. I'm getting this error again:
```
warning: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__memory/unique_ptr.h:88:5: error: aligned deallocation function of type 'void (void *, std::align_val_t) noexcept' is only available on macOS 10.14 or newer
```
This is the issue with aligned allocation/deallocation in https://github.com/iotaledger/wallet.rs/issues/1635 (it's actually supported in macOS 10.13 and Clang is incorrectly giving an error). I emailed Apple Developer Support about this in December and I don't believe I received a response. I may try posting on their forums because some of the employees are usually pretty responsive. Setting `-faligned-new` seems to silence the error but I'm not sure what happens if we do that

rajivshah3: > I'm not sure why it's not happening with the Python bindings on CI

I'm seeing similar errors when building the Python bindings locally, so I think something is cached on CI with the newer macOS target

Tuditi: As far as I understood this flag changes the way dynamical objects aligned in the memory. Enabling it instructs the compiler to automatically align dynamically allocated objects to the maximum alignment for the target platform.  I think the worst thing that could happen, is the app becoming slightly larger. What is your opinion on trying out the flag and see whether it solves the issues people seem to be having?

thibault-martinez: Should we go ahead and merge or are we still investigating this?

Tuditi: Let's merge it. If it doesn't work, we'll add the flag to see what happens.

thibault-martinez (approved):

.github/workflows/build-and-test.yml
thibault-martinez: Python bindings seem to fail to compile

thibault-martinez:
Hi @Tuditi , can you reopen this PR on the SDK repository please ?

thibault-martinez:
Closing in favour of https://github.com/iotaledger/iota-sdk/pull/15

Issue #1733, opened 2023-01-16
Iota Library for python. Recovery wallet, which has already created by phrase working incorrect.


abcen7:
## Bug description
```python
import os
import shutil
import random

import iota_wallet

path = "static/user_wallets/test"
mnemonic = 'park target phrase cheese agent turn lunch wolf broccoli glance famous camp orient maid ribbon observe decrease cave subway possible hire puppy leader chronic'


def get(password):
    account_manager = iota_wallet.AccountManager(path)
    account_manager.set_stronghold_password(password)
    wallet = account_manager.get_account('Wallet')
    wallet.sync().execute()
    print(wallet.latest_address()["address"]["inner"])


def creating(password):
    if os.path.exists(path):
        iota_wallet.AccountManager(path).stop_background_sync()
        shutil.rmtree(path)
    account_manager = iota_wallet.AccountManager(storage_path=path)
    account_manager.set_stronghold_password(password)
    account_manager.store_mnemonic("Stronghold", mnemonic)
    client_options = {"nodes": [
        {
            "url": "https://api.lb-0.h.chrysalis-devnet.iota.cafe/",
            "auth": None,
            "disabled": False
        }
    ], "local_pow": True}
    account_initializer = account_manager.create_account(client_options)
    account_initializer.alias('Wallet')
    account: iota_wallet.AccountHandle = account_initializer.initialise()
    account.sync().execute()


password = "1234"
for i in range(3):
    password += random.choice("1234")
    creating(password)
    get(password)
    print(f"Get {i}")
```

## Rust version

No version, boys

- Rust version:

No version, boys

Which version of the library are you using?

- Version number, commit, or branch: [macos-iota-wallet-py3.9-wheel](https://nightly.link/iotaledger/wallet.rs/workflows/python_binding_publish/production/macos-iota-wallet-py3.9-wheel)

## Hardware specification

What hardware are you using?

Macbook air M2 ⚠️⚠️⚠️⚠️

- Operating system:
- RAM: 8gb
- Cores: 8
- Device: Macbook air M2 

## Steps To reproduce the bug

Explain how the maintainer can reproduce the bug.

1. Download the lib, install that on your computer
2. Copy this code and paste to .py file
3. Execute and get error

## Expected behaviour

In the directory in which the wallet has already been created, delete the folder with files and create a new wallet by mnemonic

## Actual behaviour

Drops the random error 

## Errors

[...]

PR #1930, opened 2023-03-07
Add lock_stronghold and destroy methods to Python binding

Thoralf-M:
# Description of change

Add lock_stronghold and destroy methods to Python binding

Fixes https://github.com/iotaledger/wallet.rs/issues/1733

## Type of change

Added example

## Change checklist


bindings/python/examples/destroy.py
kwek20, line 1: I understand we have this for testing the deletion, but is keeping it an examplereally needed?

Thoralf-M: What speaks against having an example for this? I think it's better to have it than not

kwek20: it feels very specific to the bindings glue. if its something that should always be done it should be automated. This example doesnt actually directly use the methods we added either. 

And if were actually testing account getting etc, we should do that in the main crate

Thoralf-M: It uses `.destroy()` and recreates the account manager, the case that didn't work in the issue, but I agree that it's not really general shown and also specific to use cases where you need to recreate it
Should I change it so it just creates the account manager, destroys it and then recreates it, without anything else?

bindings/python/native/tests/test_account_manager.py
```diff
@@ -195,32 +195,32 @@ def test_account_manager_stop_background_sync():
...
-    except ValueError as e:
-        assert 'Input value is < dust protection value' in str(e)
-    else:
-        try:
-            manager.send_migration_bundle(
-                [legacy_node], bundle['bundle_hash'], min_weight_magnitude)
-
-        except ValueError as e:
-            assert 'input list is empty' in str(e)
+# def test_get_migration_data():
```
kwek20: We want this test entirely commented out? Sort of what git history is for :)

Thoralf-M: Will remove, this stuff should never be touched anymore anyways

kwek20 (approved):

thibault-martinez (approved):

PR #1951, opened 2023-03-13
documentation: updating wasm-web bindings example

l4rot:
# Description of change

I noticed that the readme for the wasm-web bindings directory was off and updated it accordingly.

Be sure to reference any related issues by adding `fixes #issue_number`.

## Type of change

- Documentation Fix

- [ x] I have made corresponding changes to the documentation

### DIFF
--- a/wallet/bindings/wasm/README.md
+++ b/wallet/bindings/wasm/README.md
@@ -141,12 +141,12 @@ plugins: [
 ### Web Usage
 
 ```javascript
-import init, { AccountManager, CoinType } from "@iota/wallet-wasm/web";
+import * as wallet from "@iota/wallet-wasm/web";
 
-init().then(() => {
-  const manager = new AccountManager({
+wallet.init().then(() => {
+  const manager = new wallet.AccountManager({
         storagePath: './my-database',
-        coinType: CoinType.Shimmer,
+        coinType: wallet.CoinType.Shimmer,
         clientOptions: {
             nodes: ['https://api.testnet.shimmer.network'],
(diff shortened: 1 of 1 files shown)


Thoralf-M (commented):
Thanks for the PR, but I think the example import looks better and we should rather change the library to work like that, what do you think?
To fix it, one probably only needs to add `\nexport default init;` here https://github.com/iotaledger/wallet.rs/blob/dc48003b762bc997b991d0fe6a33bcd08e133005/wallet/bindings/wasm/build_scripts/web.js#L63

thibault-martinez:
Hi @l4rot, we have moved to the https://github.com/iotaledger/iota-sdk repository. Please comment on Thoralf's answer and let us know what you think or reopen it on the SDK. In absence of response, we will close this PR in a week.

l4rot:
> Thanks for the PR, but I think the example import looks better and we should rather change the library to work like that, what do you think? To fix it, one probably only needs to add `\nexport default init;` here
> 
> https://github.com/iotaledger/wallet.rs/blob/dc48003b762bc997b991d0fe6a33bcd08e133005/wallet/bindings/wasm/build_scripts/web.js#L63

Thank you for your patience! 
Sounds good to me, I was just fighting with that issue for some hours until I looked up the other libs and figured it out through their approaches (identity wasm and client wasm). 
I thought this would be easier on new devs that try make their way into the iota ecosystem since it was a little bit confusing to me. 

Maybe another option would be to standardize the examples for the wasm implementations?

PR #1958, opened 2023-03-14
Consistent storage set and get

thibault-martinez:
Set was generic over a `Serialize` but get wasn't generic over a `Deserialize` so we were always deserialising after.


wallet/src/storage/manager.rs
```diff
@@ -50,10 +50,7 @@ pub(crate) async fn new_storage_manager(
         encryption_key,
     };
     // Get the db version or set it
-    let db_schema_version = storage.get(DATABASE_SCHEMA_VERSION_KEY).await?;
-    if let Some(db_schema_version) = db_schema_version {
-        let db_schema_version = u8::from_str(&db_schema_version)
-            .map_err(|_| crate::Error::Storage("invalid db_schema_version".to_string()))?;
+    if let Some(db_schema_version) = storage.get::<u8>(DATABASE_SCHEMA_VERSION_KEY).await? {
```
Thoralf-M, line 52: Should the map_err() be removed?

wallet/src/storage/participation.rs
kwek20, line 1: Do you actually need to specify the type parameter? i think if the variable has a type it can be inferred, and looks a bit cleaner

thibault-martinez: I double checked them all and one could be removed, all the other are necessary. But this is not new, the type information was already required when deserialising before.

wallet/src/storage/manager.rs
thibault-martinez: We used to `u8::from_str` externally but now we `serde_json::from_str` internally. Would be weird to map only this one and not the others ?

kwek20 (dismissed):

wallet/src/storage/manager.rs
Thoralf-M: Ah yeah it's also optional, so should be fine

Thoralf-M (dismissed):

kwek20 (approved):

Thoralf-M (approved):

PR #2000, opened 2023-08-01
Update stronghold

Thoralf-M:
# Description of change

Update stronghold to v3

Running tests and the example

## Change checklist


src/account_manager.rs
```diff
@@ -97,9 +100,15 @@ enum ManagerStorage {
 
 fn storage_password_to_encryption_key(password: &str) -> [u8; 32] {
```
semenov-vladyslav, line 104: This is really weird. We've just fixed this same issue in stronghold, and it's still there. It's not safe to derive strong encryption keys from weak passwords using fixed salt and low iteration count. This definitely must be fixed.

src/account_manager.rs
```diff
@@ -395,13 +404,15 @@ impl Drop for AccountManager {
 
 #[cfg(feature = "stronghold")]
 fn stronghold_password<P: Into<String>>(password: P) -> Vec<u8> {
```
semenov-vladyslav: Why `P: Into<String>`? Why not `P: AsRef<[u8]>` or `password: &[u8]`? Why not return `Zeroizing<Vec<u8>>`?

src/account_manager.rs
```diff
@@ -395,13 +404,15 @@ impl Drop for AccountManager {
...
-    let mut password = password.into();
-    let mut dk = [0; 64];
-    // safe to unwrap because rounds > 0
-    crypto::keys::pbkdf::PBKDF2_HMAC_SHA512(password.as_bytes(), b"wallet.rs", 100, &mut dk).unwrap();
+    let mut digest = crypto::hashes::blake2b::Blake2b256::new();
+    let mut password: String = password.into();
+
+    digest.update(password.as_bytes());
+    let key = digest.finalize().to_vec();
```
semenov-vladyslav: This is not a key, ie. it has the same very low entropy as password. Maybe rename to `password_hash` or something to avoid confusion?

src/account_manager.rs
```diff
@@ -1112,7 +1123,7 @@ impl AccountManager {
             .map_err(|_| crate::Error::PoisonError)?
             .as_ref()
         {
-            if generated_mnemonic != mnemonic.as_ref() {
+            if generated_mnemonic.deref().deref() != mnemonic.deref() {
```
semenov-vladyslav: This is very suspicious. Secrets should not be compared using standard equality, it is not constant time and may leak secret info. It seems that the logic behind this piece of code is faulty.

semenov-vladyslav: What is the purpose of `generated_mnemonic`? If it must be valid with respect to `wordlist::ENGLISH` dictionary, then why not keep the initial entropy that `generated_mnemonic` encodes? This way you can use `bip39::wordlist::decode` to get the entropy corresponding to `mnemonic` and compare entropies in constant time (using subtle crate, for example).

src/actor/mod.rs
```diff
@@ -160,17 +161,18 @@ impl WalletMessageHandler {
             MessageType::GenerateMnemonic => convert_panics(|| {
                 self.account_manager
                     .generate_mnemonic()
-                    .map(ResponseType::GeneratedMnemonic)
+                    .map(|mnemonic| ResponseType::GeneratedMnemonic((**mnemonic).to_owned()))
```
semenov-vladyslav: This leaks mnemonic into value of type `String` which should be zeroized. `ResponseType` does not implement `ZeroizeOnDrop`.

src/stronghold/mod.rs
```diff
@@ -279,20 +286,44 @@ async fn get_password(snapshot_path: &Path) -> Result<Arc<Password>> {
...
+    /// Stronghold snapshot password not set.
     #[error("snapshot password not set")]
     PasswordNotSet,
+    /// Invalid address or account index.
     #[error("invalid address or account index {0}")]
     TryFromInt(#[from] TryFromIntError),
+    /// The derived seed of a mnemonic was already stored.
     #[error("the mnemonic was already stored")]
     MnemonicAlreadyStored,
+    #[error("Stronghold migration error: {0}")]
```
semenov-vladyslav: I suppose error messages should be lowercase

src/stronghold/mod.rs
```diff
@@ -279,20 +286,44 @@ async fn get_password(snapshot_path: &Path) -> Result<Arc<Password>> {
...
+    /// Stronghold migration error.
+    Migration(#[from] iota_stronghold::engine::snapshot::migration::Error),
+    /// Invalid number of hash rounds.
+    #[error("invalid number of hash rounds: {0}")]
+    InvalidRounds(u32),
+    /// Path already exists.
+    #[error("path already exists: {0}")]
+    PathAlreadyExists(std::path::PathBuf),
+    /// Unsupported stronghold snapshot version, migration required.
+    #[error("Unsupported snapshot version, migration required")]
```
semenov-vladyslav: Here too.

src/stronghold/mod.rs
```diff
@@ -579,13 +626,16 @@ pub async fn generate_address(
     check_snapshot(&mut runtime, snapshot_path, None).await?;
     load_private_data_actor(&mut runtime, snapshot_path, None).await?;
 
-    let chain = Chain::from_u32_hardened(vec![
+    let chain = [
```
semenov-vladyslav, line 636: Why not `use crypto::keys::bip44`? This way you can use `bip44::Bip44::from([4218, account_index, internal, address_index])` to represent BIP44 chain. It's easier and safer to use, you don't have to manually `.into_iter().map(|s| Segment::harden(s).into()).collect()`.

semenov-vladyslav (changes requested):
`storage_password_to_encryption_key` thing must definitely be fixed. There's no easy way to derive strong key from weak password and virtually no salt and iteration count.

src/account_manager.rs
Thoralf-M: This comparison is actually pretty stupid to have at all

semenov-vladyslav:
Regarding `storage_password_to_encryption_key`. As @Thoralf-M explained, the key derived with it is used to encrypt public data that doesn't need thorough protection (integrity guarantees provided by the inner DB would suffice). It would be good idea to get rid of this function and respective code in the future.

src/stronghold/mod.rs
Thoralf-M: Hm but Stronghold requires `type SLIP10Chain = Vec<u32>` and if I use `Bip44::.to_chain::<ed25519::SecretKey>()` I get this mismatch errror
```
mismatched types
expected struct `Vec<u32>`
    found array `[crypto::keys::slip10::Hardened; 5]
```

Thoralf-M: Hm but Stronghold requires `type SLIP10Chain = Vec<u32>` and if I use `Bip44::.to_chain::<ed25519::SecretKey>()` I get this mismatch errror
```
mismatched types
expected struct `Vec<u32>`
    found array `[crypto::keys::slip10::Hardened; 5]
```

src/stronghold/mod.rs
Thoralf-M: Changed

src/actor/mod.rs
Thoralf-M: Changed `Response::GeneratedMnemonic()` to have `Zeroizing<String>`

src/account_manager.rs
Thoralf-M: Added a comment why it's like this

src/account_manager.rs
Thoralf-M: If it's a ref we can't zeroize it :thinking: 
Changed the return type

src/stronghold/mod.rs
Thoralf-M: Changed

src/account_manager.rs
semenov-vladyslav: But you are zeroizing the value of type `String`, not the initial value of type `P`. `P` should transfer ownership to password in `into` method, so I guess that works. Not sure what the right way to do it.

src/stronghold/mod.rs
semenov-vladyslav: Right. We should have used Bip44 in stronghold Slip10 API. With `Vec<u32>` it is more generic but less convenient.

src/stronghold/mod.rs
```diff
@@ -3,13 +3,21 @@
...
-use crypto::hashes::{blake2b::Blake2b256, Digest};
-
-use crypto::keys::slip10::Chain;
+mod migration;
+
+use crypto::{
+    hashes::{blake2b::Blake2b256, Digest},
+    keys::{
+        bip39::{Mnemonic, Passphrase},
+        bip44::Bip44,
```
semenov-vladyslav: Unused import.

src/stronghold/mod.rs
```diff
@@ -27,11 +35,11 @@ use tokio::{
     sync::Mutex,
     time::{sleep, Duration},
 };
-use zeroize::Zeroize;
+use zeroize::{Zeroize, Zeroizing};
 
 #[derive(PartialEq, Eq, Zeroize)]
 #[zeroize(drop)]
```
semenov-vladyslav: Use `derive(ZeroizeOnDrop)` instead of `zeroize(drop)`.

src/stronghold/mod.rs
```diff
@@ -253,7 +261,7 @@ fn default_password_store() -> Arc<Mutex<HashMap<PathBuf, Arc<Password>>>> {
     Default::default()
 }
 
-pub async fn set_password<S: AsRef<Path>>(snapshot_path: S, password: Vec<u8>) {
+pub async fn set_password<S: AsRef<Path>>(snapshot_path: S, password: Zeroizing<Vec<u8>>) {
```
semenov-vladyslav, line 262: What is the deal with password types? In one place it's `P: Into<String>`, this mod defines `struct Password`, and here it is `Zeroizing<Vec<u8>>`.

semenov-vladyslav: [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed) makes a good point that passwords/strings should be normalized as the different GUI libraries can represent the same chars with different unicode codepoints which will result in different byte slices. `crypto::keys::bip39::Passphrase` is a safe type to use in this case. Maybe we should make is standalone definition in crypto independent of bip39 mod.

semenov-vladyslav (changes requested):
A few minor changes required. Type of passwords should be unified.

src/stronghold/mod.rs
Thoralf-M: The `Password` is not for BIP39, it's only internally for the stronghold module (it's `pub`, but the stronghold module is private)

src/account_manager.rs
Thoralf-M: Checked and `<P: Into<String>>` is actually from `AccountManager::change_stronghold_password()`, changing this would be a breaking change, therefore will not change this

src/stronghold/mod.rs
Thoralf-M: It's updated for the new stronghold version, but will not happen in this version and wallet.rs anymore.

